PR #53, opened 2014-02-14
Adding basic types for OpenCV

mtourne:
Some of the basic types for OpenCV
The mat type is binary and won't archive well with non-binary archivers

OpenCV's FileStorage api still might be a bit more convenient to use (supplying matrixes by hand)
But this can be useful.

### DIFF
--- a/include/cereal/types/opencv.hpp
+++ b/include/cereal/types/opencv.hpp
@@ -1,0 +1,114 @@
+/* Copyright (c) 2014 Matthieu Tourne */
+
+#ifndef CEREAL_TYPES_OPENCV_HPP_
+#define CEREAL_TYPES_OPENCV_HPP_
+
+#include <cereal/cereal.hpp>
+#include <opencv2/opencv.hpp>
+
+namespace cereal {
+
+    // cv::Mat
+    // this is largely inspired from opencv's persistence.cpp:icvWriteMat
+    template<class Archive>
+    void save(Archive &ar, const cv::Mat& m) {
+        int elem_type = m.type();
+        size_t step = m.step;
+        size_t elem_size = m.elemSize();
+        size_t elem_count = m.total();
+        int y;
+
+        ar( _CEREAL_NVP("rows", m.rows) );
+        ar( _CEREAL_NVP("cols", m.cols) );
+
+        ar( _CEREAL_NVP("elem_type", elem_type) );
+
+        // type number and size of elements
+        ar( _CEREAL_NVP("elem_size", elem_size) );
+        ar( make_size_tag( elem_count) );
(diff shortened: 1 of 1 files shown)


patrikhuber:
I also wrote cereal serialisation functions for OpenCV a few months ago and I have a few remarks regarding this PR:

This pulls in all kinds of headers from OpenCV for `cv::DMatch` and `cv::KeyPoint`. These should be separated from what is in OpenCV `core` (i.e. `cv::Mat`, `cv::Vec`, etc.). Ideally we'd have a cereal module for each of the OpenCV modules that require serialisation, but I think it's enough to just serialise the types in `core` for a start. Certainly we shouldn't require the user to pull in modules others than `core`, so these additional types should be removed from the PR.

OpenCV has a proxy type for its data structures, `OutputArray` and `InputArray`. I'm not sure whether they can or should be used, but we should think about it. (one reason against it would be that for example `std::vector<cv::Vec3f>` is an `InputArray` as well, but in this case we wouldn't want OpenCV to serialise it but rather use `cereal/types/vector.hpp` to serialise the `std::vector` and our OpenCV serialisation to serialise the `cv::Vec3f`'s).

The serialisation for `cv::Mat` in this PR is not as simple as it could be - it's sufficient to serialise  `mat.rows`, `mat.cols`, `mat.type()` and `mat.isContinuous()`. You can find my implementation in my blog: http://www.patrikhuber.ch/blog/6-serialising-opencv-matrices-using-boost-and-cereal which in my opinion is much simpler.

Seems to me like you shouldn't use `CV_IS_MAT_CONT` (deprecated C-API macro).

Also, in my opinion, supporting XML serialisation for `cv::Mat` doesn't make sense, because the files get very large very quickly, and if you look at them, each matrix entry will be inside its own XML attribute, so it's not really "human-readable" anyway. So I would remove `_CEREAL_NVP`. (But in case you want to keep it, I think you should use `cereal::make_nvp` instead).

Interestingly, I believe in your implementation if you save a non-continuous matrix, it will have `continuous = true` after loading it. We should think about whether we want this behaviour.

include/cereal/types/opencv.hpp
```diff
@@ -0,0 +1,114 @@
...
+        int queryIdx, trainIdx, imgIdx;
+        float distance;
+
+        ar( queryIdx, trainIdx, imgIdx, distance);
+
+        match = cv::DMatch(queryIdx, trainIdx, imgIdx, distance);
+    }
+}
+
+#endif // !CEREAL_TYPOS_OPENCV_HPP_
```
smvv: I assume TYPOS should be TYPES?

PR #66, opened 2014-02-27
Added a 'getNodeName' API to JSON/XML InputArchives

mattyclarkson:
If you are doing some pretty custom serialization/deserialization it can really help being able to know the current node key/tag.

I couldn't really figure out where to put this into the unit tests - any hints? Would you want tests for this?

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -558,6 +558,12 @@ namespace cereal
         ++itsIteratorStack.back();
       }
 
+      //! Retrieves the current node name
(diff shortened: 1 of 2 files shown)


mattyclarkson:
@AzothAmmo, did you actually put this into the repo? I can't find the function in the cereal releases?

AzothAmmo:
This is strange, I definitely remember adding this but it is also definitely not there.  Looking through the commits to see what happened.

mattyclarkson:
I was doing the same thing - it's in the wiki pages but not in the code. Couldn't work out what had happened.

PR #122, opened 2014-09-14
Add support for boost::optional

stevehickman:
Copyright notice is identical to your original wording, with my name there instead.
This should be sufficient to support boost::optional

### DIFF
--- a/unittests/common.hpp
+++ b/unittests/common.hpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2014, Randolph Voorhies, Shane Grant
+  Copyright (c) 2014, Randolph Voorhies, Shane Grant, Steve Hickman
   All rights reserved.
 
   Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


thelink2012:
It seems to be missing move semantics in some places, such as in the boost::optional load():

    optional = val;

could be

    optional = std::move(val);

Anyway pretty cool job you are doing on supporting the boost types!

include/cereal/types/boost/compressed_pair.hpp
```diff
@@ -0,0 +1,62 @@
...
+  //! Saving for boost::compressed_pair
+  template <class Archive, class T1, class T2> inline
+  void save( Archive & ar, ::boost::compressed_pair<T1, T2> const & pair )
+  {
+    ar( _CEREAL_NVP("first",  pair.first()),
+        _CEREAL_NVP("second", pair.second()) );
+  }
+  
+
+   //! Loading for boost::optional
```
patrikhuber: I think this should say "boost::compressed_pair", not optional

include/cereal/types/boost/filesystem.hpp
```diff
@@ -0,0 +1,130 @@
+/*! \file boost_optional.hpp
+\brief Support for boost::optional
```
patrikhuber: I think this should say filesystem, not optional :-)

patrikhuber:
Not being able to serialise `boost::optional` bit me today. Do you think this could be merged and integrated into the next release, at least the part about `boost::optional`? I think optional is probably the most widely used type of all these.

AzothAmmo:
It is something I want to do in the near future but I want to separate out serialization code for external libraries from the core of cereal, so it won't make the next immediate release but is targeted for the one after that.

patrikhuber:
@AzothAmmo I do agree with you. However, don't you think `boost::optional` specifically is worthy of being included directly? It's pretty much "core functionality", widely used and will probably even make it into the standard soon. I see boost_variant in the included types as well ;-) But I see your point, I can also just copy the header from here over into my project.

AzothAmmo:
Well we can always bring it in and then push it out to a different place once we do modules.

patrikhuber:
I would (of course) leave the decision up to you. I can offer to prepare a PR. I just figured however since I include the cereal headers in my projects directly (and not via submodule or external module or something), it actually doesn't really make a difference and I can just include the file for optional in my projects. I'm OK with either.

stevehickman:
Per instructions on the web site, I've now made this and other updates on the current develop branch and submitted a new pull request. See PR #282

joshhaines:
It would be awesome to get this merged into the next version. If there's anything I can do to help, let me know. I had to fork Cereal and pull in this boost stuff (along with a tweak to get Cereal working with Android). However, it would be awesome if this was officially supported.

stevehickman:
I'll get the boost support synched with the latest develop branch this weekend.

PR #141, opened 2014-12-14
allow for cereal to be applied to template classes with multiple arguments

avimosher:
This uses __VA_ARGS__ instead of T in the macro argument list to allow for template classes with multiple arguments.  I _think_ it should be strictly more functionality and not break anything that worked before (but I might be wrong).  Also, I didn't touch CEREAL_REGISTER_TYPE_WITH_NAME because I would have had to switch the argument order in the macro to use __VA_ARGS__, which would definitely break things.


AzothAmmo:
Original PR was from a while ago so it was easier to do this one by hand.

PR #142, opened 2014-12-20
Implement minimal JSON Print

erichkeane:
The JSON archive uses the RapidJSON::PrettyWriter, which inserts
a bunch of extra whitespace.  This change adds an option to use
the RapidJSON::Writer, which generates a minimum-length string.

Signed-off-by: Erich Keane erich.keane@verizon.net


chrisdonlan87:
Was there ever a solution to this?

Issue #140, opened 2014-11-12
Tuple creates json file with multiple same-named keys


rggjan:
Wher serializing a tuple:

    std::tuple<float, char, std::string> tuple(3.8, 'A', "Lisa Simpson");
    oar( cereal::make_nvp("best data ever",  tuple) );

with the string serializer, you get output like this:

    {
        "best data ever": {
            "tuple_element": "Lisa Simpson",
            "tuple_element": 65,
            "tuple_element": 3.7999999523162842
        }
    }

This is bad, because parsing it might be difficult (although it works with cereal):
http://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object/23195243#23195243

Maybe we should change it to "tuple_element0, tuple_element1, ...), same as the values in an object are numerated?

PR #143, opened 2014-12-20
Implement Tuple Name Differences in XML/JSON

erichkeane:
User rggjan requested that tuples provide a different name for each
element in this issue: https://github.com/USCiLab/cereal/issues/140.

This fix modifies tuple.hpp to generate tuple names in the order which
they are are placed in the tuple.  Note that this will appear to be a
countdown (tuple_element2, tuple_element1, tuple_element0) since tuples
are generated in reverse order.

Signed-off-by: Erich Keane <erich.keane@verizon.net>

### DIFF
--- a/include/cereal/types/tuple.hpp
+++ b/include/cereal/types/tuple.hpp
@@ -45,7 +45,9 @@ namespace cereal
       template <class Archive, class ... Types> inline
       static void apply( Archive & ar, std::tuple<Types...> & tuple )
       {
-        ar( _CEREAL_NVP("tuple_element", std::get<Height - 1>( tuple )) );
+        
+        ar( _CEREAL_NVP(("tuple_element" + std::to_string(Height -1)).c_str(),
+          std::get<Height - 1>( tuple )) );
         serialize<Height - 1>::template apply( ar, tuple );
       }
     };


erichkeane:
Any comments or anything on this one? Should i remove the ordering change?

AzothAmmo:
Does this actually change the ordering?  It looks like it is still counting down instead of counting up.  Other than that it looks good - I do want to change the order to be in the natural tuple order (0, 1, 2, ...)

erichkeane:
It DOES reverse it.  Note how I've reversed the call to apply and the ar( call.

This is my test snippet:

    std::tuple<int, int, int, int> tup(1,2,3,4);
    ostringstream os;
    {
        cereal::JSONOutputArchive ar(os);
        ar(tup);
    }
    std::cout<< os.str()<<std::endl;


With apply happening second, this is the output:

    {
        "value0": {
            "tuple_element3": 4,
            "tuple_element2": 3,
            "tuple_element1": 2,
            "tuple_element0": 1
        }
    }
However, with the apply second (like in 58d1675) this is my output:

    {
        "value0": {
            "tuple_element0": 1,
            "tuple_element1": 2,
            "tuple_element2": 3,
            "tuple_element3": 4
        }
    }

AzothAmmo:
Right you are.  My eyes glossed over the unchanged line in the diff.  Will fully test and merge this into 1.1.

AzothAmmo:
Everything seems fine on Linux, I don't expect any problems on Windows.  Will compile with MSVC later anyway, but closing now.

PR #152, opened 2015-01-08
Fix missing return value keyword in access class

HungMingWu:


javenPHD:
Curious, were in access class is keyword missing? Is there an example?

AzothAmmo:
This is already fixed on the develop branch, see 9addb12616144d9f0b81e0516e80550b4633c3f2

javenPHD:
Awesome. Ty for the heads up.

PR #156, opened 2015-01-12
Json and Xml archive fixes

ChrisBFX:
There was a buffer over-read when enabling the output of the types in the xml outputarchive.
Creating a JSONOutputArchive without serializing anything results in a failed assertion when the destructor was called.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -153,7 +153,8 @@ namespace cereal
       //! Destructor, flushes the JSON
       ~JSONOutputArchive()
       {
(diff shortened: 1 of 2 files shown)


erichkeane:
Commit 1 was fixed differently in #155 which has been pulled into the dev repo.  IMO, the 2nd patch is really nice, it is annoying that it crashes like this.

AzothAmmo:
We'll take the second patch for sure, thanks for the contribution.

AzothAmmo:
Merged to develop (9fb72cf6c419e745c6dfaabe134fc34083498fcf)

Issue #109, opened 2014-07-28
XML deserialization for fully whitespaced strings loads empty strings


volo-zyko:
When I deserialize the following XML "&lt;value&gt;  &lt;/value&gt;" to std::string I get empty string. This happens because RapidXML ignores whitespaces. This could be fixed with "&lt;value xml:space='preserve'&gt;  &lt;/value&gt;" during serialization but this is not supported by RapidXML during parsing. What RapidXML supports is CDATA. So, during serialization std::strings could be saved as "&lt;value&gt;&lt;![CDATA[  ]]&gt;&lt;/value&gt;" and RapidXML can parse this construct.

Issue #107, opened 2014-06-26
VS2013 problem when polymorphic serialization classes exist in a library


runarundheim:
I created a test with two polymorphic classes. If the classes and the write code exist in the same project everything works nice. If I try to split up the polymorphic classes in a library and link this library into the project where the serialization is performed then the code fail.
The CEREAL_REGISTER_TYPE doesn't seem to register when the class is registered in a library.
I also get this warning:

warning C4505: 'cereal::detail::polymorphic_serialization_supportcereal::BinaryInputArchive,T::instantiate' : unreferenced local function has been removed

Is there a way to force this code so that it is not removed when the library is created?

robiwano:
I'm using v1.1.2, but there are no CEREAL_REGISTER_SHARED_LIBRARY  etc. in polymorphic.hpp ?? Please advise.

Issue #105, opened 2014-06-24
Long serialization ambiguous in VS on develop branch


AzothAmmo:
This is related to fixes introduced by #81.  See relevant discussion [here](https://groups.google.com/forum/#!topic/cerealcpp/df29MDDVtsE) on the mailing list.

Issue #81, opened 2014-03-24
Bitset serialization from JSON fails to compile  with clang/libc++ on OS X


randvoorhies:
This is the standard `unsigned long` =/= `uint64_t`, and `long` =/= `int64_t` issue:

```
/Users/rand/workspace/cereal/./include/cereal/archives/json.hpp:599:12: note: candidate function not viable: no known conversion from 'unsigned long' to
  'uint64_t &' (aka 'unsigned long long &') for 1st argument
```

This can be fixed by adding an extra case for `long` and `unsigned long`. I will commit these changes in a moment.

PR #157, opened 2015-01-12
Concrete archives being templates plus some minor improvements.

ImplOfAnImpl:
Hi.
In this pull request the archive classes that were concrete classes previously have become templates that take a Derived class as a parameter, which is then passed to OutputArchive/InputArchive. This allows one to extend an existing archive's functionality by inheriting from it in CRTP manner. The original archives' names exist as typedefs, so users won't notice the change, except for slightlly bigger build times.

Also some issues have been fixed here:
1. JSON archive's compilation problems on Mac
2. XML archive's problem of serializing space characters and space-only strings.

And some tweaks have been made:
1. CEREAL_REGISTER_TYPE can now be called from multiple translation units for the same type. This allows one to re-register an already registered type when additional archives become available.
2. Error message has been improved for the case when multiple serialization routines exist for the same type.
3. CEREAL_DECLARE_DYNAMIC_INIT_ENFORCER and CEREAL_DEFINE_DYNAMIC_INIT_ENFORCER macros have been added, which allow to force the compiler to initialize global data of a certain translation unit. 
I personally found the need for them when I had a source file that called CEREAL_REGISTER_TYPE while it and its associated header file didn't have anything that is "odr-usable" - without the macros I had the "Trying to save/load an unregistered polymorphic type" exception on every recompile/run. A colleague of mine had a more subtle issue, where the source file with CEREAL_REGISTER_TYPE contained some definitions for the hierarchy's base class' members but not for the constructor, which was defined as inline in the header file; the application he was working on was only deserializing data that came through a network connection; this lead to the "unregistered polymorphic type" error appearing on every other recompile/run, probably becase the constructor was put to that translation unit by the compiler sometimes, which made it "odr-used" before the deserialization for that type had begun.
Note that in practice I've only seen this issue when the serializable type and the CEREAL_REGISTER_TYPE call were in a static library and the serialization was performed outside of it, this is why the corresponding tests use static libraries.

### DIFF
--- a/include/cereal/archives/binary.hpp
+++ b/include/cereal/archives/binary.hpp
@@ -48,15 +48,15 @@ namespace cereal
       inadvertently.
 
       \ingroup Archives */
-  template <class Derived, std::uint32_t Flags = 0>
-  class BinaryOutputArchiveT : public OutputArchive<Derived, Flags | AllowEmptyClassElision>, public detail::BinaryArchiveBase
+  template <class Derived>
+  class BinaryOutputArchiveT : public OutputArchive<Derived, AllowEmptyClassElision>, public detail::BinaryArchiveBase
   {
     public:
       //! Construct, outputting to the provided stream
       /*! @param stream The stream to output to.  Can be a stringstream, a file stream, or
                         even cout! */
       BinaryOutputArchiveT(Derived * derived, std::ostream & stream) :
-        OutputArchive<Derived, Flags | AllowEmptyClassElision>(derived),
+        OutputArchive<Derived, AllowEmptyClassElision>(derived),
         itsStream(stream)
       {
         static_assert(std::is_base_of<BinaryOutputArchiveT, Derived>::value, "The passed class must derive from this one");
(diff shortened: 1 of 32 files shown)


AzothAmmo:
This will take a little while to process.  Some of your changes may already have been performed in our develop branch, but there are some interesting things here, especially everything related to `REGISTER_TYPE`.

AzothAmmo:
Going to check off various things that I've found we've already pushed to develop bit by bit.

Your b5a520e7dbfafd1b9b5435e1fe415ff92c6acecb (static assertion improvement) was done by 2edc936e5006a01e8ed33854040e0d9147d45c3d.

AzothAmmo:
XML space saving was resolved by #109 (not sure which of yours does this).

Your `CEREAL_DECLARE_DYNAMIC_INIT_ENFORCER` was implemented by e423228dde132a1be238096491b9c1e36dadfb2d (see #107).

JSON compilation issues have likely been fixed by a series of changes to long serialization on OSX (see #81 and #105).

I haven't looked into the CRTP changes yet.  However, moving prologue/epilogue functions to be member functions is likely problematic for future user-defined types since they would require direct modification of the archive itself.

Your changes to the registration for polymorphic items are going to solve two issues that have been blocking 1.1 for a while.

ImplOfAnImpl:
>> XML space saving was resolved by #109 (not sure which of yours does this).

It's funny, the guy who posted this is one of my colleagues, and he sits a couple of meters away from me. It looks like we have a communication problem :-)

>> Your CEREAL_DECLARE_DYNAMIC_INIT_ENFORCER was implemented by e423228 (see #107).

OK. Just a note: the problem is not specific to shared libraries (it's reproducible for me with static ones) and in theory it's not related to libraries at all - the compiler is just not required to initialize global data from a translation unit until something from it is actually used. Also, CEREAL_REGISTER_SHARED_LIBRARY must not only be in the library, it must be in the same translation unit where CEREAL_REGISTER_TYPE is put (I checked this with Clang and GCC on Ubuntu).  So, it's worth to rename the macro and update the comment.

>> JSON compilation issues have likely been fixed by a series of changes to long serialization on OSX (see #81 and #105).

I see. Though my solution is a bit cleaner IMO.

>> I haven't looked into the CRTP changes yet. However, moving prologue/epilogue functions to be member functions is likely problematic for future user-defined types since they would require direct modification of the archive itself.

The default implementations of member prologue/epilogue delegate to the non-member ones, so user-defined types will still be able to have custom overloads for them.

AzothAmmo:
Had a chance to give archive-templates a better looking over and will likely include with 1.2.  I had toyed with a solution when trying to solve #46 that did something similar but ultimately went with something more light weight but less flexible.

M2tM:
Awesome, I'll check it out.

PR #184, opened 2015-04-18
Added support for std::valarray.

ilerik:
Motivation
I'm using cereal++ as underlying serialization lib for high perfomance scientific computing, and std::valarray is the type specifically held in STL for this particular purpose. I found that std::valarray serialization wasn't supported yet, so i decided to contribute to this nice project with my implementation. Please be kind to review this pull request.

Here is what has beed done :
1) Implemented std::valarray serialization (based on vector.hpp) in file valarray.hpp
2) Implemented unit tests in file valarray.cpp (based on vector.cpp tests, except for bool case)
3) vs2013/unittests project configuration updated accordingly
4) unittests/common.hpp updated.
5) No documentation so far (not sure how to do it properly)

### DIFF
--- a/include/cereal/types/valarray.hpp
+++ b/include/cereal/types/valarray.hpp
@@ -1,0 +1,103 @@
+/*! \file valarray.hpp
+\brief Support for types found in \<valarray\>
+\ingroup STLSupport */
+
+/*
+Copyright (c) 2014, Randolph Voorhies, Shane Grant
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


AzothAmmo:
Looks pretty good overall - will give this a detailed look and likely integrate into 1.1.2.

As far as documentation goes don't worry about it - we'll just have to rebuild doxygen and coverage.

ilerik:
Glad to hear that. C++11 is just so good and classy) And cereal++ is a very good example of some very nice code done C++11 way.

PR #199, opened 2015-06-16
Fix shared_ptr costum deleter

ph03:
Correctly prevent calling destructor on unitialized data

### DIFF
--- a/include/cereal/types/memory.hpp
+++ b/include/cereal/types/memory.hpp
@@ -271,7 +271,7 @@ namespace cereal
       ptr.reset( reinterpret_cast<T *>( new ST() ),
           [=]( T * t )
           {
-            if( valid )
+            if( *valid )
               t->~T();
 
             delete reinterpret_cast<ST *>( t );


erichkeane:
1- I think this patch is a good one that fixes a bug.

2- I wonder if the shared_ptr<bool> is even necessary?  It seems to me that it is only used for thread-safety, so perhaps a std::atomic<bool> would be more efficient?

ph03:
I was thinking that it's just to prevent calling the destructor on uninitialized memory (which happened to me even in single threaded mode). But I guess you are right, the lambda could be swapped with a deleter functor class having an atomic bool indicating if deletion is valid / save, which is set after successful object construction.

erichkeane:
I'm not the maintainer here or anything, or the author, so I just saw it and decided to comment. My opinion isn't really worth all that much  here:)

I was saying that the full lambda capture (which is unnecessary btw) is a waste, and using a shared_ptr for its atomicy (ensuring it doesn't skip deleting if it has finished) is a waste.  My suggestion on this one would be to have the lambda capture ONLY valid by ref, and make 'valid' be a std::atomic, since it won't require a mutex or a deref anymore.

randvoorhies:
@AzothAmmo will have to elaborate, but it looks like the ```shared_ptr<bool> valid``` business was added in 3a92f0bb to help with loading circular references, but I haven't had time to look at it in depth.

This fix appears straightforward to me though, so I'm cool to accept it. @AzothAmmo?

AzothAmmo:
Yeah no problem accepting this.  I feel like there was some reason I have a `shared_ptr` in there instead of just a reference to a bool (atomic or not), but I can't remember what my thought process was.  It's entirely possible it isn't necessary at all.

ph03:
Do you still consider merging this for the sake of fixing the bug or do you rather want to overhaul this part of the code?

AzothAmmo:
Merging it, will potentially revisit any code optimization at another time.

PR #205, opened 2015-07-14
Generalize cereal/types/map.hpp to deal with arbitrary pair associative containers having mapped_type.

abutcher-gh:
This allows any container having mapped_type be serialized by the map serializer.

Since the map serializer is now more generally specified, the patch also has the effect of allowing more specialization options without introducing ambiguity ([example here](https://github.com/abutcher-gh/cereal/commit/f4c973d61101c94df319e6bb86f80ce4c4112742#diff-3b3781f71a23082ce2d849059eec7b88R834)).


AzothAmmo:
I like the look of this!

erichkeane:
I've been looking at this too, I'm a pretty big fan of abstracting out the need for specifically mentioning the map.  Also gives us a few of the alternate map types (including boost!).  Seems like a solid pull request to me.

AzothAmmo:
I'm going to merge this shortly after I make a decision on the following:

Existing users probably expect to be able to include cereal type support for something in the standard library without having to include the header for the type itself.  So I see two options moving forward here:

1. Remove the `<map>` and `<unordered_map>` includes as you have done.  This is technically a breaking change and will run contrary to the behavior of all other type includes. 
2. Leave the `map` and `unordered_map` includes in place, but introduce a third header, maybe called `<map_concept>`, which avoids these includes for those not needing them.

Thoughts?

abutcher-gh:
What about <cereal/concepts/map.hpp> for this?  Then have <cereal/types/map.hpp> and <cereal/types/unordered_map.hpp> include it and their respective std container header.

abutcher-gh:
Maybe having <cereal/concepts/pair_associative_container.hpp>

abutcher-gh:
Closed by mistake.  Reopened now.  I was going to say, maybe the concept header should be `<cereal/concepts/pair_associative_container.hpp>`.  Having said that, I only find the AssociativeContainer concept at http://en.cppreference.com/w/cpp/concept/AssociativeContainer with a note that `map` and `multimap` additionally define `mapped_type`.

abutcher-gh:
I pushed a sketch.  What do you think?

AzothAmmo:
I like the idea of a concepts directory for things like this.  I'll merge this shortly and make a note to update any related documentation.

Issue #82, opened 2014-03-24
Update to recent RapidJSON and integrate streaming (was JSON-parser errors are very unhelpfull)


Fiona-J-W:
The error-messages in the exceptions thrown by the JSON-parser aren't very helpfull:

> Error: rapidjson internal assertion failure: IsObject()

It would be great if these could become more descriptive, maybe something like this:

> Error: in object 'value0': in array 'value3': expected object, but found integer.

ajneu:
> **randvoorhies** commented on Mar 24, 2014
>
> It would be great if we could figure out how to add **line numbers** to errors as well. Ideally, the above error would read something like:
>
> Cereal JSON serialization error. Expected type [string] on **line 4**, but found [list] instead.

Jip, **line numbers** would be really good. 
Is there a way of doing this currently? 
(Perhaps getting some kind of marker of stream position or something? Then I could count newlines myself if I have a stream that allows me to go back to the beginning.)

Thanks.

<sub>**Side note: Regarding JSON and C++11:**</sub>
<sub> * [JSON Voorhees](https://github.com/tgockel/json-voorhees)</sub>
<sub> * [JSON for Modern C++](https://github.com/nlohmann/json)</sub>

<sub>(from https://isocpp.org/search/google?q=json )</sub>

AzothAmmo:
We're now using RapidJSON 1.0.2, I haven't done anything other than make it work as is, so no streaming or anything like that yet.

correaa:
I also found this problem  when adding nvp information (value is a custom type)

This worked,
```
	cereal::JSONOutputArchive{os}<< value;
```

then
```
	cereal::JSONOutputArchive{os}<< CEREAL_NVP(value); // or cereal::make_nvp("value", value);
```
gave the "rapidjson internal assertion failure: IsObject()" error.

Then I tried 
```
	cereal::JSONOutputArchive{os}<< cereal::make_nvp("value0", value);
```
and it worked again.

I tried with strings like `"valu0"` or `"val0"` or `"valuex"` or `"value1"` and they all fail as well. 
But `"value00"` worked. It seems that there is something special about finishing with a number "0".

Also, in all the failing cases, I see an extra "`}`" at the end of the file.

PR #221, opened 2015-08-20
Current rapidjson

m7thon:
EDIT Feb 17, 2016 (update to newest cereal and rapidjson):

This updates rapidjson to the current master and discards the modifications made by cereal. The differences to the previously included version of rapidjson are, as far as I can tell:

* The `precision` for double output has a different meaning. Currently rapidjson supports "truncated" double output, e.g., a precision of 2 means: 0.1285 -> 0.12, 0.004 -> 0.0, 1.23 e-7 -> 0.0. 

* I modified the current rapidjson to support writing and parsing of `nan`, `inf` and `-inf` doubles. This is a non-standard json extension that rapidjson currently does not support (but may in the future). Note that the ability to parsing `-inf` seems to be a new feature for cereal.

* `long double`, `long long`, and `unsigned long long` are no longer handled in rapidjson. However, it seems this is now handled by cereal anyway by writing these types as strings.

The file `json.hpp` is modified only minimally to work with the current rapidjson. Importantly, the parser is given the `rapidjson::kParseFullPrecisionFlag`, which means that double values are parsed in full precision (AKA correctly). This ensures that the roundtrip C++ -> json -> C++ works for all doubles.


AzothAmmo:
I'll look at this in the near future since I have a little more time for cereal right now - looks like automated tests are failing but I would sort this out in a merge.  This was on my list of things to do (see #82) so if this looks good and integrates well that'd be a great help.

m7thon:
Compilation just gives one harmless `-Wshadow` error in rapidjson/reader.h line 752. You'll have to see how you want to deal with it.

Disabling `-Werror` as I have done above is probably not what you want to do ;-)

m7thon:
By the way, it seems your CI build doesn't actually run your unit tests, just compiles them. Is that intended?

AzothAmmo:
They get run after compilation.

-----Original Message-----

bagobor:
'rapidjson/filewritestream.h' - file is absent in latest rapidjson packege.

m7thon:
@bagobor Do you mean filestream.h ? This has been replaced by filereadstream.h and filewritestream.h, and I think these now work slightly differently. You can have a look at the rapidjson docs [here](https://github.com/miloyip/rapidjson/blob/master/doc/stream.md#file-streams-filestreams).

But this (filestream.h) was not used by cereal anyway.

bagobor:
@m7thon I'm very sorry ( was incorrect 'paste'), I meant file 'rapidjson/genericstream.h'

m7thon:
Ah yes. But you can use the old file. I updated it to match the example given in the current rapidjson documentation, and use the RAPIDJSON... namespace macros.

Issue #218, opened 2015-08-11
Doc: 'version' type inconsistency


patrikhuber:
On http://uscilab.github.io/cereal/transition_from_boost.html, `version` is of type `const unsigned int`, probably because that's the type boost uses. Cereal seems to use `std::uint32_t const` (http://uscilab.github.io/cereal/serialization_functions.html), so there's an inconsistency here in the documentation. I think at least in the later examples on the first page where the _preferred cereal style_ is shown, the type should be changed to `const unsigned int`. Not sure how/if to change the first example where boost and cereal are compared in the same example.

If you want, I can submit a PR with the changes. In that case, let me know how you want to proceed with the first example.

AzothAmmo:
Good catch - we should be advocating use of the sized type (`std::uint32_t`) over `unsigned int`.  If you want to do a PR and I haven't already changed this, change all of the cereal versions to use `std::uint32_t`.  I'll add a comment in afterwards emphasizing this change to users.

patrikhuber:
Done, see #227. As requested, I left it up to you to add a comment. I think [L70](https://github.com/USCiLab/cereal/pull/227/files#diff-365eb3eaa8c059ebf7c03d4dae0021d6R70) or around [L138](https://github.com/USCiLab/cereal/pull/227/files#diff-365eb3eaa8c059ebf7c03d4dae0021d6R70) might be good places to add something.

PR #227, opened 2015-09-03
Doc: Changed version parameter to std::uint32_t const

patrikhuber:
This is to be consistent with the rest of the documentation. See #218.


AzothAmmo:
merged into gh-pages-develop

PR #236, opened 2015-09-28
Save bitset as binary data.

lyager:
- Use a shifting bitmask to serialize bits.
 - Only use bitset type, others are just as slow.

Converting bitsets to strings by using 1's and 0's takes up too much space on large bitsets.

### DIFF
--- a/include/cereal/types/bitset.hpp
+++ b/include/cereal/types/bitset.hpp
@@ -41,9 +41,7 @@ namespace cereal
     /*! @internal */
     enum class type : uint8_t
     {
-      ulong,
-      ullong,
-      string
+      bits
     };
   }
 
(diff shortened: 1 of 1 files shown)


AzothAmmo:
This has the beginnings of a good change though we'll need to make some adjustments before merging something like this.  Serializing like this is great for binary archives but wouldn't look good in a text based archive.  In other situations like this (see vector for example) we use the `BinaryData` wrapper to create two versions of the serialization code - one for archives optimized to handle binary chunks, and one for everyone else.

So that would be the first thing that needs to be done.

The second thing is that all of the types need to be the fixed sized variants, especially things that actually go to be serialized (`uint8` instead of `char` for example).

These are pretty minor requirements so if you don't get around to making them yourself, I'll do them when I do the merge.  Otherwise it's a good idea!

lyager:
Thanks for the constructive feedback.. I agree, and I will try to submit a new patch soon.

AzothAmmo:
I've integrated your changes such that archives that support `BinaryData` will now automatically save using your chunking method. When loading it will always dynamically check the bitset serialization enum type so that this is backwards compatible.

Issue #217, opened 2015-08-06
MSVC JSON: mix serialize and save_minimal error


sschaetz:
Hi everyone,

I ran into a runtime issue with the JSON archives on visual studio express 2013. It seems like I can not mix `serialize` and `save_minimal` functions when serializing hierarchical types. Here is a reproducer:

```cpp
#include <stdio.h>
#include <stdint.h>
#include <cereal/archives/json.hpp>

template <typename T>
struct foo
{
	T value;
};


template<class Archive, typename T>
T save_minimal(const Archive&, const foo<T>& v)
{
	return v.value;
}

/*
// This version works.
template<class Archive, typename T>
void serialize(Archive & archive, foo<T>& v)
{
	archive(v.value);
}
*/

struct bar
{
	foo<uint8_t> v1;
	foo<uint8_t> v2;
	foo<uint16_t> v3;
	foo<uint8_t> v4;
};


template<class Archive>
void serialize(Archive & archive, bar& m)
{
	archive(::cereal::make_nvp("v1", m.v1));
	archive(::cereal::make_nvp("v2", m.v2));
	archive(::cereal::make_nvp("v3", m.v3));
	archive(::cereal::make_nvp("v4", m.v4));
}


int main(void)
{
	bar b;
	// Serialize to stream.
	std::ostringstream out;
	{
		cereal::JSONOutputArchive outArchive(out);
		outArchive(::cereal::make_nvp("param", b));
	}
}
```

This crashes when writing the third value. I started debugging and it seems like it is not considering the fact that the values serialized with save_minimal don't have a name. In `writeName` the first line popping the stack seems to fail because the stack is zero. The portable binary archive seems to work.

I hope somebody can reproduce this and help fix the issue.

Best,
Sebastian

sschaetz:
The same code works fine on Linux with gcc 4.8.3.

AzothAmmo:
I'll take a look at this - can confirm it works correctly on Linux.

sschaetz:
Azoth, thanks for looking into this. Did you make any progress - could you reproduce the problem?

AzothAmmo:
I can reproduce this on MSVC2013.

AzothAmmo:
Seems that the minimal serialization is incorrectly entering the non-minimal [epilogue](https://github.com/USCiLab/cereal/blob/master/include/cereal/archives/json.hpp#L781) for `foo<unsigned char>`.

[...]

PR #240, opened 2015-10-08
Additional MSVC JSON minimal fixes

kersson:
Hey guys, the fixes in f451090 still did not solve USCiLab/cereal#217 when I tested it using VS Express 2013 for Desktop with Update 5 (msvc-12.0).

I changed the SFINAE logic to use `EnableIf` with AND conjunctions via the variadic arguments instead of `DisableIf` with the buggy OR conjunctions.  This fixed the issue for me and @sschaetz's example code now runs without issue.  To me this decreases the readability of the code, but at least it now works! :smile:

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -753,40 +753,40 @@ namespace cereal
       that may be given data by the type about to be archived
 
       Minimal types do not start or finish nodes */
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Looks ugly but it's better than chasing quirks with MSVC forever.

PR #244, opened 2015-11-06
Failure case for XML deserialization of strings ending with whitespaces

BigSeb:
Fixed an issue with the preservation of whitespaces for strings. The whitespace validation was invalid in a particular case. Since the itsOS stream is always being reused for the same archive, data coming from previous save operation was enlarging the stream buffer. When calling itsOS.str() this copy not only the first part of the data being serialized (itsOS << value << std::ends) but the rest of the buffer is also copied to the new string (strValue). We may end up on a string that contains multiple null terminating characters. The validation for the last whitespace character was done on the last character of the string which may be data coming from previous serialization. The validation may then fails if that left over does not contain a whitespace even if the current value to be serialized contains a whitespace.

The test case for reproducing the failure case is simple. Simply save a long string at first (no whitespace at the end) and then save a shorter string that contains string at the end. The "xml:space=preserve" attribute is not added for the second attribute even though it should.

I have added a test that reproduce this error which was mostly based on previous test.

The fix I propose shrinks the strValue string so that it only contains the newly serialized value up to the first null character. The next calls to length() is then matching where the first null character is.

### DIFF
--- a/include/cereal/archives/xml.hpp
+++ b/include/cereal/archives/xml.hpp
@@ -232,19 +232,22 @@ namespace cereal
         itsOS.clear(); itsOS.seekp( 0, std::ios::beg );
         itsOS << value << std::ends;
 
-        const auto strValue = itsOS.str();
+        auto strValue = itsOS.str();
+
+        // itsOS.str() may contains data from previous calls after the first '\0' that was just inserted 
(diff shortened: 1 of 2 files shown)


AzothAmmo:
Will try and roll this in when I have time to look at everything, most likely in December.

PR #256, opened 2016-01-16
Added named variant serialization

klemens-morgenstern:
## Idea

This feature allows variants to be serialized with their typename not the index identifying the variant. 

This makes the XML and JSON files look nice, but also includes the string into the binary archives (I have no idea to how that could be avoided).

### The easiest way is to use boost::type_index to determine the typename. This may however make the names not portable.

```cpp
boost::variant<int, double> var1 = 42;
ar & CEREAL_NVP_NAMED_VARIANT(var1);
```

Which would yield the following result:

```json
{
   "var1" : { "type" : "int", "data" : 42 }
}
```

Now one can use *CEREAL_NVP_NAMED_VARIANT_* to give an explicit name instead of *var1* similar to *CEREAL_NVP_*.

Alternatively one can give the matching names like this:

```cpp
boost::variant<int, double> var = 2.0;
ar & CEREAL_NVP_NAMED_VARIANT_("my_variant", var, "integer", "floating");
```

Which will yield:

```json
{
   "my_variant" : { "type" : "floating", "data" : 2.0" }
}
```

## Unit-Test

I also provided a unit test, which passes. This addition does however need at least boost 1.56 because it uses boost/type_index to determine the type names automatically. This causes the current test to fail.

## Added Dependency

  - [boost.type_index](http://www.boost.org/doc/libs/1_60_0/libs/preprocessor/doc/index.html)
  - [boost.preprocessor](http://www.boost.org/doc/libs/1_60_0/doc/html/boost_typeindex.html)


fiesh:
I don't think this will work for `std::variant` which allows the same types to be present multiple times.  

That being said, we actually do have serializations similar to that, but they are handled on a per-archive basis.  I don't think this should be changed globally.

klemens-morgenstern:
Well it would work, just take the first intance when deserializing. Or index them. Would be a different solution, is there an interest?

fiesh:
> Well it would work, just take the first intance when deserializing. Or index them. Would be a different solution, is there an interest?

Taking the first occurence would lead to the very surprising behavior that serializing a variant `v` and deserializing it into a variant `w` will result in `v != w`.

PR #260, opened 2016-02-13
Make make_nvp member of archives

sschaetz:
There are instances where, in a given compilation unit, we never want to serialize a cereal-serializable type but only use it. 

However, if the type has a serialize() member that uses ::cereal::make_nvp, the cereal
header must  be included.

We propose adding make_nvp to the archives themselves so a serializable type is usable without including any cereal headers. But at the same time it will still be possible to maintain the serialization code as a member of the type, which is preferable over external definition of the serialization code in a separate file.

The modified unit-test make_nvp_member.cpp illustrates the idea: It is now possible to implement the type `unordered_naming` without the type knowing anything about cereal. To achieve this the archive concept is extended with the make_nvp method.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -157,6 +157,23 @@ namespace cereal
           itsWriter.EndObject();
       }
 
+      // ######################################################################
+      //! Creates a name value pair
+      /*! @relates NameValuePair */
+      template <class T> static inline
+      NameValuePair<T> make_nvp( std::string const & name, T && value )
+      {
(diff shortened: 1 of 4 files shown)


sschaetz:
I just realized I created a pull request for master.  Sorry about that.  Should I recreate a pull-request for a feature branch?

AzothAmmo:
It is a lot easier for us to potentially integrate changes if done vs develop.

sschaetz:
Got it, created https://github.com/USCiLab/cereal/pull/261 
What did you think of the proposal (-> respond in other pull request)?

PR #261, opened 2016-02-13
Make make_nvp member of archives

sschaetz:
There are instances where, in a given compilation unit, we never want to serialize a cereal-serializable type but only use it.

However, if the type has a serialize() member that uses ::cereal::make_nvp, the cereal
header must be included.

We propose adding make_nvp to the archives themselves so a serializable type is usable without including any cereal headers. But at the same time it will still be possible to maintain the serialization code as a member of the type, which is preferable over external definition of the serialization code in a separate file.

The modified unit-test make_nvp_member.cpp illustrates the idea: It is now possible to implement the type unordered_naming without the type knowing anything about cereal. To achieve this the archive concept is extended with the make_nvp method.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -162,6 +162,23 @@ namespace cereal
           itsWriter.EndArray();
       }
 
+      // ######################################################################
+      //! Creates a name value pair
+      /*! @relates NameValuePair */
+      template <class T> static inline
+      NameValuePair<T> make_nvp( std::string const & name, T && value )
+      {
+        return {name.c_str(), std::forward<T>(value)};
+      }
+
+      //! Creates a name value pair
+      /*! @relates NameValuePair */
+      template <class T> static inline
+      NameValuePair<T> make_nvp( const char * name, T && value )
+      {
+        return {name, std::forward<T>(value)};
+      }
+
       //! Saves some binary data, encoded as a base64 string, with an optional name
       /*! This will create a new node, optionally named, and insert a value that consists of
           the data encoded as a base64 string */
@@ -550,6 +567,23 @@ namespace cereal
       }
 
     public:
+      // ######################################################################
+      //! Creates a name value pair
+      /*! @relates NameValuePair */
+      template <class T> static inline
+      NameValuePair<T> make_nvp( std::string const & name, T && value )
+      {
+        return {name.c_str(), std::forward<T>(value)};
+      }
+
+      //! Creates a name value pair
+      /*! @relates NameValuePair */
+      template <class T> static inline
+      NameValuePair<T> make_nvp( const char * name, T && value )
+      {
+        return {name, std::forward<T>(value)};
+      }
+
       //! Starts a new node, going into its proper iterator
       /*! This places an iterator for the next node to be parsed onto the iterator stack.  If the next
           node is an array, this will be a value iterator, otherwise it will be a member iterator.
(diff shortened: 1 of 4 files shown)


erichkeane:
Rather than copy and pasting the same forward declarations into each of the serialization headers, would it be better to pull these functions off into their own forward-declaration-like-header (as boost does often with these sorts of things), and have those be included instead?  

That way you wouldn't need to include the entirety of json.hpp, and instead could just include cereal_fwd_decl.hpp or something.

Thoughts?

sschaetz:
Two points:
I also am not a big fan of copy-pasting the make_nvp function into the archives. If somebody has an idea to keep this DRY, I'll implement it.

Having a `cereal_fwd_decl.hpp` only helps to a certain degree however: 

What a user really wants it to implement types that are serializable and not specify the library that is used to serialize the types. This way any Archive that implements the function-call operator and the `make_nvp` method can be used. With a forward header, types are still tied to cereal as they call `::cerleal::make_nvp()` and include `cereal_fwd_decl.hpp`.

erichkeane:
That said, requiring an Archive to be defined (and for those types to KNOW of their archive) break a big part of the purpose of separating out the Archives.  There is no reason that type MyFoo should need to know of an archive at all. 

I suspect that common non-archiving (meaning they don't actually DO the serialization) should not have to know anything about the archives list.  I'd suggest pulling out functions commonly used in those contexts into their own file.

My cereal_fwd_decl.hpp was a suggestion for a place, since Boost tends to put a bunch full-on template implementations in their fwd_decl headers for exactly these cases.

sschaetz:
Exactly my point. `MyFoo` should know nothing about serialization if serialization is not required. That is the beauty of making the `make_nvp` method part of the Archive. *Only if the archive is used, i.e. serialization is performed* must anything be known about serialization.

`MyFoo` does not require the archive to be defined. It does not now and it will not with this change. Without this change, an archive must assume that `Archive` implements `operator()`, now it must assume it also implements `make_nvp`. No difference.

Check out this snippet to emphasize my point. http://coliru.stacked-crooked.com/a/644894973dcd2a20

This works because the method is a template.

AzothAmmo:
I haven't had much time to think about this yet, so these are just initial thoughts and I will consider this. My gut reaction is that I don't like seeing the same code repeated, and I'm a bit worried about other similar features (e.g. base class casting, `BinaryData`, etc) having similar requests made to be part of an archive.

sschaetz:
I understand those concerns. I'm happy to explore ways to avoid repeating the same code. Do you have any ideas? The only thing I can think of right now is giving Archives an additional type that, by default, has the `make_nvp` method. Then there would be only one type that has `make_nvp` and all Archives would be constructed (default?) with this. So `archive has-a helper` and `helper has make_nvp`. 

This could also serve as a customization points for archives. Users can expose more functionality through the archive to be usable at the point of serialization. Not sure what one would want to put in there though as the current design is quite elegant and complete. Except for the missing `make_nvp`.

If I understand your second point, you're worried this might open a can of worms of more requests to add more stuff to archives. That is certainly a concern but I don't see how this change request changes that. Are you worried this might set a precedence? In that case it might help to thing of this not as an enhancement but as a bug-fix. The fact that types with member serialization that make use of `make_nvp` require inclusion of a cereal header is a bug.

sschaetz:
Any new thoughts on this?

sschaetz:
Bump

PR #266, opened 2016-03-08
Fix memory leak due to alloc-free size mismatch

tdudziak:
That fixes a memory bug I found while running my cereal-using code with [AddressSanitizer](https://github.com/google/sanitizers). I would add a testcase but the error seems to appear only with AddressSanitizer (Valgrind doesn't report it).

[Here](https://gist.github.com/tdudziak/8694e47444c47a26f516) is a simple program and the output it produces when compiled with clang with `-fsanitize=address`. The problem is basically that `aligned_storage<sizeof(T)>` uses default alignment (which might be different from `T`'s alignment) and thus might add additional padding.

This is most likely not a big deal and I don't think it actually leaks this extra memory but technically it relies on undefined behavior (I think) and it might annoy users of the library that test their code with AddressSanitizer.

### DIFF
--- a/include/cereal/types/memory.hpp
+++ b/include/cereal/types/memory.hpp
@@ -119,7 +119,7 @@ namespace cereal
       // typedefs for parent type and storage type
       using BaseType = typename ::cereal::traits::get_shared_from_this_base<T>::type;
       using ParentType = std::enable_shared_from_this<BaseType>;
-      using StorageType = typename std::aligned_storage<sizeof(ParentType)>::type;
+      using StorageType = typename std::aligned_storage<sizeof(ParentType), alignof(ParentType)>::type;
 
       public:
         //! Saves the state of some type inheriting from enable_shared_from_this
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Not sure why the travis build failed it doesn't look related to your changes. `alignas` is not supported in GCC 4.7 which is still our minimum supported version, so we need to have a workaround for that compiler to accept this change.

tdudziak:
I'm only using `alignof`, not `alignas`. Both are part of N2341 which [isn't officially supported](https://gcc.gnu.org/gcc-4.7/cxx0x_status.html) on GCC 4.7 but `alignof` seems to work. I checked on Debian Wheezy with GCC 4.7.2-5.

Still, I can produce a workaround if you prefer that.

AzothAmmo:
Ah woops, I misread. Everything is fine then! I'll make sure this doesn't trip up anything and then merge it.

PR #277, opened 2016-04-22
Implementation and testing of std::atomic serialization.

bluescarni:
This is a simple implementation of serialization for `std::atomic` types via their load/store methods.

Although typically used with integral types, `std::atomic` can be instantiated with any type that is trivially copyable. The current implementation requires `T` to be default-constructible for the load method.


jossgray:
Could overload prologue and epilogue for JSON / XML archives to prevent an extra node being created.

Issue #280, opened 2016-04-27
Tests unittests/map.cpp compile error under VS2013/VS2015 with boost 1.6X


stevehickman:
VS apparently does not define stream operators for std::pair. Compiling map.cpp results in:
  ```
map.cpp
1>d:\boost\boost_1_61_0_b1\boost\test\tools\detail\print_helper.hpp(47): error C2338: Type has to implement operator<< to be printable
1>  d:\boost\boost_1_61_0_b1\boost\test\tools\detail\print_helper.hpp(178): note: see reference to class template instantiation 'boost::test_tools::tt_detail::print_log_value<T>' being compiled
1>          with
1>          [
1>              T=std::pair<const std::string,int>
1>          ]
1>  d:\boost\boost_1_61_0_b1\boost\test\utils\wrap_stringstream.hpp(66): note: see reference to function template instantiation 'std::ostream &boost::test_tools::tt_detail::operator <<<std::pair<const _Kty,_Ty>>(std::ostream &,const boost::test_tools::tt_detail::print_helper_t<std::pair<const _Kty,_Ty>> &)' being compiled
1>          with
1>          [
1>              _Kty=std::string,
1>              _Ty=int
1>          ]
1>  d:\boost\boost_1_61_0_b1\boost\test\tools\old\impl.hpp(213): note: see reference to function template instantiation 'boost::basic_wrap_stringstream<char> &boost::operator <<<char,boost::test_tools::tt_detail::print_helper_t<std::pair<const _Kty,_Ty>>>(boost::basic_wrap_stringstream<char> &,const T &)' being compiled
1>          with
1>          [
1>              _Kty=std::string,
1>              _Ty=int,
1>              T=boost::test_tools::tt_detail::print_helper_t<std::pair<const std::string,int>>
1>          ]

```
One way to eliminate this problem is to add the following to the top of map.cpp:

```
typedef  std::pair<const std::string, int>       str_int_map_type;
BOOST_TEST_DONT_PRINT_LOG_VALUE(StructBase)
BOOST_TEST_DONT_PRINT_LOG_VALUE(str_int_map_type)

```
See this `[link  ](http://www.boost.org/doc/libs/1_60_0/libs/test/doc/html/boost_test/test_output/testing_tool_output_disable.html)`for documentation.

stevehickman:
Made a typo on the solution. Add the following to map.cpp:

typedef  std::pair<const std::string, int>       str_int_map_type;
[...]

PR #282, opened 2016-04-30
Develop

stevehickman:
Added a boost subdirectory under include/cereal/types and unittests with support for:

boost::chrono
boost::circular_buffer
boost::compressed_pair
boost::dynamic_bitset
boost::filesystem
boost::optional
boost::units

includes update to unittests/common.hpp to pull in these additional headers

Also includes a VS2015 solution that complements the VS2013 solution. The only distinction is that the VS2015 solution includes $(BOOST_HOME) in the additional include dirs and $(BOOST_LIBRARYDIR) in the additional library dirs (because that's how my system is set up)


AzothAmmo:
Looks really nice. I'm going to take care of #280 for this next immediate release and clean up any remaining problems on VS related to updates for #82, but I'll make the focus of 1.3 integrating new features such as this.

PR #284, opened 2016-05-06
added virtual destructors to ArchiveBase classes

reuk:
Currently, if you call `delete` on a pointer to InputArchiveBase or OutputArchiveBase, you'll leak the contents of the derived classes because the base classes have non-virtual destructors.

Might not be an actual issue in practice, but I get compiler warnings unless I add the virtual destructors, and I don't like warnings!

### DIFF
--- a/include/cereal/details/helpers.hpp
+++ b/include/cereal/details/helpers.hpp
@@ -225,16 +225,26 @@ namespace cereal
     /* The rtti virtual function only exists to enable an archive to
        be used in a polymorphic fashion, if necessary.  See the
        archive adapters for an example of this */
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
:+1:

AzothAmmo:
I'm cool with adding this but we only accept pull requests from the develop branch. This is a really minimal change so I can take care of it unless you want to re-submit a PR.

reuk:
Thinking about it, defining a destructor will stop the compiler auto-generating move assigns/constructs.

I'll submit a pull to the develop branch with a full complement of `default` move/copy assigns/constructs.

PR #285, opened 2016-05-17
Make `double` roundtrip when serializing and deserializing to JSON

m7thon:
The current develop branch of cereal has several problems with serialization and deserialization of `double` values:

1. Values are **truncated** at 17 decimal digits (this is **not** the same as having 17 decimal digits of *precision*). For example, `1e-20` is serialized as `0.0`. The setting to avoid truncation is given by `rapidjson::Writer<>::kDefaultMaxDecimalPlaces` and is equal to 324 decimal placed (not 17). I think you may have misunderstood the `SetMaxDecimalPlaces()` option of the current rapidjson writer.

2. The special values `nan`, `inf` and `-inf` are not serialized by the current version of rapidjson (nothing is printed at all, resulting in an invalid json representation that cannot be deserialized, i.e., crashes on deserialization).

3. With the current settings (without `rapidjson::kParseFullPrecisionFlag`), the deserialization loses up to 3 bits of precision. For example, deserializing `0.9868011474609375` becomes `0.9868011474609376`.

This pull request presents a possible fix to all of these issues (1 and 3 being rather trivial changes to json.hpp)

By the way, note that the version of rapidjson you currently include in the develop branch is not 1.0.2, but a rather recent version from the master branch.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -108,7 +108,7 @@ namespace cereal
           static Options Default(){ return Options(); }
 
           //! Default options with no indentation
-          static Options NoIndent(){ return Options( std::numeric_limits<double>::max_digits10, IndentChar::space, 0 ); }
+          static Options NoIndent(){ return Options( JSONWriter::kDefaultMaxDecimalPlaces, IndentChar::space, 0 ); }
 
           //! The character to use for indenting
           enum class IndentChar : char
@@ -124,7 +124,7 @@ namespace cereal
               @param indentChar The type of character to indent with
               @param indentLength The number of indentChar to use for indentation
                              (0 corresponds to no indentation) */
-          explicit Options( int precision = std::numeric_limits<double>::max_digits10,
+          explicit Options( int precision = JSONWriter::kDefaultMaxDecimalPlaces,
                             IndentChar indentChar = IndentChar::space,
                             unsigned int indentLength = 4 ) :
             itsPrecision( precision ),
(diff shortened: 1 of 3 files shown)


AzothAmmo:
I'll merge this shortly. Have you considered submitting e5911ad as a PR to rapidjson as well? I still need to modify rapidjson to play nicely with the MSVC warnings and will likely see if I can get that merged upstream as well.

m7thon:
I submitted a modified [PR](https://github.com/miloyip/rapidjson/pull/641) to rapidjson that adds the functionality of writing/parsing "NaN", "Inf", and "-Inf". This plays more nicely with other parsing options of rapidjson.

m7thon:
Ok, so writing and reading of nan/inf/-inf as "NaN", "Inf" and "-Inf" is now possible with rapidjson as an option.

Use by substituting `rapidjson::PrettyWriter<WriteStream>` with `rapidjson::PrettyWriter<WriteStream, rapidjson::UTF8<>, rapidjson::UTF8<>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag>` and the call `itsDocument.ParseStream<rapidjson::kParseFullPrecisionFlag>(itsReadStream);` with `itsDocument.ParseStream<rapidjson::kParseFullPrecisionFlag | rapidjson::kParseNanAndInfFlag>(itsReadStream);`

Alternatively, one can also change the default setting by macros. Before including rapidjson, define:

```
#define CEREAL_RAPIDJSON_WRITE_DEFAULT_FLAGS kWriteNanAndInfFlag
#define CEREAL_RAPIDJSON_PARSE_DEFAULT_FLAGS kParseFullPrecisionFlag | kParseNanAndInfFlag
```

Then, don't specify any flags for reading and parsing.

PR #297, opened 2016-06-06
added virtual destructor to PolymorphicCaster

reuk:
I found another struct with no virtual destructor.


AzothAmmo:
Note: MSVC2013 doesn't support default implementations for move ctors/assignment or noexcept. I'll have to make sure our NOEXCEPT macro is usable throughout cereal.

reuk:
That concern will also apply to my previous pull request #287 then.

AzothAmmo:
I took care of it after the merge, will do the same with this one.

PR #307, opened 2016-06-27
Provide manual implementations for to_string() & friends for Android NDK + GCC

rcdailey:
By defining `CEREAL_IMPLEMENT_TO_STRING` via the preprocessor, self-implemented
versions of the following STL functions will be provided.

* `std::strtold()`
* `std::to_string()`
* `std::stoi()`
* `std::stol()`
* `std::stoll()`
* `std::stoul()`
* `std::stoull()`
* `std::stof()`
* `std::stod()`
* `std::stold()`

The specific use case for this is using Android NDK with C++11 or higher
and on the GCC 4.9 + GNU STL toolchain/library. At the moment, implementations
of these libraries are not provided by this toolchain.

### DIFF
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -39,6 +39,13 @@
 #include <cstdint>
 #include <functional>
 
+// Allow a custom implementation of std::to_string, std::stol, and other
+// functions. As an example for when this is needed: GCC in Android NDK
+// does not implement std::to_string.
+#if defined(CEREAL_IMPLEMENT_TO_STRING)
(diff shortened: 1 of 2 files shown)


rcdailey:
After giving this a little more thought, I'm going to close this PR.

Ultimately this is a toolchain problem and not the responsibility of Cereal to solve. Other libraries might try to implement their own versions as well, so this might cause ambiguity/conflicts later.

I'm going to restructure this code outside of cereal in my personal code base and make CMake transparently include `to_string.hpp` so that everything using `std::to_string()` doesn't know the difference. It will appear to come from STL directly.

Issue #315, opened 2016-07-13
Cereal Thread Safe for different archives?


volcoma:
So is it thread safe to load whole different archives on different threads. Is something preventing cereal to  operate multiple archives in parallel? Please somene answer as i did not find apropriate answer for this

ChrisBFX:
No, sadly it is not. Cereal uses several StaticObjects which are not thread safe. For example the versioning-feature uses a static std::unorderd_map without a mutex.

AzothAmmo:
We don't have the mutexes in place but it is easy to make cereal thread safe for cases like you describe by adding mutexes over the global objects. I believe there have been prior discussions on how to do this. Perhaps in a future release we can make a preprocessor option to compile in locks on these operations.

AzothAmmo:
The current develop branch should be thread safe so long as each archive is only accessed by one thread, though you will need to ensure `CEREAL_THREAD_SAFE=1` is defined (see cereal/macros.hpp). Documentation will follow shortly.

volcoma:
well thanks man, that's great :+1:

Issue #210, opened 2015-07-20
Is Cereal Thread Safe?


ssbanerje:
I had a question about the thread safety of `cereal`. I have the following code, which uses cereal to serialize an `int` and a `std::vector` into a `cereal::BinaryOutputArchive`. It does 10000 of such conversions in parallel.

``` c++
#include <iostream>
#include <cassert>
#include <vector>
#include <tuple>

#include <tbb/tbb.h>

#include <cereal/cereal.hpp>
#include <cereal/archives/binary.hpp>
#include <cereal/types/vector.hpp>
#include <cereal/types/tuple.hpp>

std::string hexdump(const uint8_t* const begin, const size_t nbytes) {
  std::string str{""};
  const uint8_t* end = begin + nbytes;
  str.reserve((end - begin) * 2);
  const char *hex_chars = "0123456789ABCDEF";
  uint64_t ctr = 0;
  for (const uint8_t* it=begin; it < end; ++it) {
    uint8_t tmp = *it;
    ctr = (ctr + 1) % 16;
    char hex_string[4];
    hex_string[3] = '\0'; hex_string[2] = ' ';
    hex_string[1] = hex_chars[tmp % 16];
    tmp /= 16;
    hex_string[0] = hex_chars[tmp % 16];
    str += hex_string;
    if (ctr == 0) {
      str += "\n";
    }
  }
  return str;
}

class message {
 protected:
  char* buffer_;
  size_t buffer_length_;

 public:
  message() : buffer_(nullptr), buffer_length_(0) {}

  message(size_t length) {
    buffer_length_ = length;
    buffer_ = new char[length];
  }

  ~message() {
    if (buffer_) {
      delete [] buffer_;
      buffer_ = nullptr;
    }
  }

  inline char* buffer() const { return buffer_; }

  inline void set_buffer(char* buf, const int len) { buffer_ = buf; buffer_length_ = len; }

  inline unsigned buffer_length() const { return buffer_length_; }

  inline std::string dump_message() const {
    std::stringstream sstream;
    sstream << hexdump(reinterpret_cast<uint8_t*>(buffer_), buffer_length_);
    return sstream.str();
  }
};


class message_factory {
 public:
  template <typename... Args>
  static message create_message(Args&&... args) {
    std::stringstream stream;
    cereal::BinaryOutputArchive archive(stream);
    archive(std::forward<Args>(args)...)  ;
    auto buffer = stream.str().c_str();
    const size_t buffer_length = stream.str().size();
[...]

PR #320, opened 2016-07-28
Issue 315: making cereal thread safe

ChrisBFX:
I've found two place which manipulate static objects in an unsafe manner:
`Versions::find()` in helpers.hpp
`PolymorphicVirtualCaster()` in polymorphic_impl.hpp

These are now guarded by `std::unique_lock` and `std::mutex`. These guards are enabled with the new compile define CEREAL_THREAD_SAFE or the cmake define THREAD_SAFE (default off to maintain backwards compatibility)


AzothAmmo:
Looks good at first glance, I'll try and test this out and merge asap. Do you think these are the only two places we need to add a mutex? I'll look around myself when I go to merge this.

AzothAmmo:
I'm thinking of moving the mutex locking over to the `getInstance()` function of `StaticObject`. I feel like this is the safest option to take here since these are all global objects. 

It may be the case that this causes some extra locks, mostly when cereal is doing its initial set-up, but it catches all potential shared accesses in a clean way that doesn't involve any thought when writing future code for cereal.

Another change I'm going to make is to also add your preprocessor definition to `cereal/macros.hpp`, since users may not always be using CMAKE to handle compilation.

AzothAmmo:
see discussion #315, #210

ChrisBFX:
> I'm thinking of moving the mutex locking over to the getInstance() function of StaticObject. I feel like this is the safest option to take here since these are all global objects. 

I'm not sure this is possible as this would only guard the creation and returning of the object, but would not prevent two threads changing the same instance of a StaticObject.

AzothAmmo:
You're correct. Only thing I'm debating is whether it would look nicer to have an interface that would allow something like

```cpp
auto lock = StaticObject<InputBindingMap<Archive>>::lock();
```

versus the need to have a lot of `#ifdef` in code.

ChrisBFX:
Yes, that's a lot better.

AzothAmmo:
Try out develop and let me know how this works out. I added some tests to try and capture the threading but I don't know if they do a good job; I wasn't able to create any errors with them.

ChrisBFX:
Works for me. :+1: 
The tests run without errors with the new compile flag and fail otherwise (i commented the #ifdef out in the tests to see what happens), so I guess they are good enough.

PR #321, opened 2016-07-28
Json/XML Archive improvement

ChrisBFX:
Better error messages when loading of nvp fails. You can now see the complete path of nvps that fails which helps if you're parsing big xml/json archives with nested data structures.

The name of the root node in the xml can now be configured on a per archive basis (default is the one set with the CEREAL_XML_STRING_VALUE macro, so no everything should work as before) which is easier than the macro (which is not always possible if you want to use different root names and polymorphism support).

### DIFF
--- a/unittests/xml_archive.cpp
+++ b/unittests/xml_archive.cpp
@@ -5,21 +5,20 @@
 
 // Tests the default behavior to swap bytes to current machine's endianness
 BOOST_AUTO_TEST_CASE( xml_rootnodename )
-{ 
-
-    std::ostringstream os;
-    {
-      cereal::XMLOutputArchive oar(os, cereal::XMLOutputArchive::Options(std::numeric_limits<double>::max_digits10, true, false, "xmlroot"));
-      int32_t i_int32 = 42;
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Can you clarify the NVP exception changes you've made here? We already throw an exception (`cereal::Exception`) if we can't find the NVP - are you trying to catch other potential errors (I see you catch `std::exception`).

ChrisBFX:
Yes this catches more errors, f.e. when rapidjson can't parse a value cause of a type-mismatch it throws an exception, so this would give you more info. (Though I'm seeing now you modified rapidjson to throw an cereal::Exception, so i guess it would be ok to change the catch in the json archive to cereal::Exception, the xml archive though uses the std::stoll (and similar) functions to parse the values which throw std::exception so i wouldn't change the catch there).
Another reason is that it's sometimes difficult to find out which NVP exactly is missing if you have nested data structures like:
```
{
  "id": 1,
  "config": {
     "id": 2,
     "data": {
         "identifier": 3
      }
   }
}
````

Issue #326, opened 2016-08-08
Allow cereal to assign default values to missing name value pairs


rcdailey:
I use cereal for HTTP body response parsing for JSON/XML. Sometimes, the JSON may have optional elements: They may or may not appear in the JSON or XML data. In these cases, I do not want an exception to be thrown if the corresponding NVP is missing. Instead, I would like a default value (i.e. `T{}`, where T is the type being deserialized), or be able to specify the default value used if such an element does not appear.

I realize there is already `boost::optional` support, but this is a little different: *all* members in all cereal supported objects need to be optional by default, because I don't always know in advance which will be missing (could be user error, for example, but it's not *exceptional*, and parsing should resume normally).

Any thoughts on how this change could be implemented? I'd be fine with a configurable setting that is off by default which enables "optional" parsing, or some type of specialized `CEREAL_NVP` implementation.

rcdailey:
I just realized this is also discussed in #30. However, this is long overdue. I'd like to discuss possible implementations and help with a PR. Any ideas are welcome. Hopefully this is an easy change.

erichkeane:
What I've done before is have my deserialize function try/catch around each call to archive, then do archive.setName(nullptr) in the catch.  This resets the value, and gave me a chance to set it to whatever I wished.

pwm1234:
@rcdailey says "I realize there is already boost::optional support." I have not seen anything about cereal working with boost::optional? In what files is support for boost::optional implemented? (I searched the cereal repo for optional.hpp and found no matches.)

something-generic:
@pwm1234 there are several pull requests out that address boost optional, see 276 and 122.

Issue #30, opened 2013-10-20
Optional Fields


karhu:
I was wondering if support for optional serialization fields is considered/planned.

In my case I would use this feature when loading/writing settings files from/to json (or xml), where certain entries don't have to be present. But also in the scenario of binary serialization this could lead to memory savings.

Based on the boost_compat_new branch I was able to implement this for my own maybe<T> type and the json archives. For this I had to add another variant of the templated save / load functions of the json-archive and type specific prologue/epilogue functions.

Is there any interest in officially supporting something similar?

aseering:
The discussion above seems to imply a straightforward implementation -- implement a load() method that catches and appropriately handles the exception that's thrown in case the field is missing.  Is that what y'all're thinking of doing?  Or did you have deeper changes in mind?

heinermann:
The problem with catching exceptions is that the performance will be significantly impacted. And when I write significantly, I'm not exaggerating.

aseering:
Indeed...  Is there a cheaper way to intercept this failure currently in Cereal?
On Mar 31, 2016 10:03 PM, Adam Heinermann <notifications@github.com> wrote:The problem with catching exceptions is that the performance will be significantly impacted. And when I write significantly, I'm not exaggerating.

Type1J:
The experience from Protocol Buffer, Cap'n Proto, etc. explicitly states that treating missing fields as an error is a bad thing.

However, the other side of that argument is that state that is expected to be modified on load will not be, and instead would retain its previous value. However, if the object always resets its state on load to some default state, then it wouldn't be a problem.

I'd say that it's easier to work around problems caused by having all fields optional than working around all fields being required. Because it's easier, the exception throw on not finding a field to which a deserialized value can bind should be removed, and Cereal should increment its version number.

@gramic I've been looking for a BSON Cereal archive. Have you released the BSON Cereal archive that you talked about above? If so, where could I get it?

gramic:
@Type1J I didn't open source BSON Cereal archive, mainly because I use very old versions of both BSON and Cereal libraries. Well, also some of the code is not efficient as much as it should be to show off.

Recently, MongoDB did create a very modern c++11 libraries, which will be perfect for use in a new modern version of BSON Cereal archive that I intent to use and then open source.

About the missing optional field in Cereal, my current use is to **require** every field to be present. That forces me to update all document fields in MongoDB before using the newly added field via the BSON Cereal archive. A bit inconvenient, but it works.

[...]

PR #328, opened 2016-08-08
WIP: Introduction of cereal::make_optional_nvp()

rcdailey:
Allows a default value to be assigned to name value pairs that are not found by an input archive. For an NVP representing a value of type T, the default value assigned if that NVP is missing is `T{}` or whatever was specified explicitly. Example:

```
ar(CEREAL_OPTIONAL_NVP(myfoo, "default_value"));
ar(cereal::make_optional_nvp("myfoo_different", myfoo, "default_value"));
```

This PR is a work in progress. The overall design of this needs review. Corresponding issue is #326 and #30.

Concerning design, this is a very simple solution. An extra step that should be taken is to clean up the exception throw site. The archive itself shouldn't throw, but instead use an error code. The corresponding load function for the NVP object should throw instead. This allows the optional NVP functions to not be impacted performance-wise by the throwing that is ultimately swallowed simply to set the default value.

### DIFF
--- a/include/cereal/archives/xml.hpp
+++ b/include/cereal/archives/xml.hpp
@@ -759,6 +759,30 @@ namespace cereal
   { }
 
   // ######################################################################
+  //! Prologue for NVPs for XML output archives
+  /*! NVPs do not start or finish nodes - they just set up the names */
+  template <class T> inline
+     void prologue(XMLOutputArchive &, OptionalNameValuePair<T> const &)
+  { }
+
+  //! Prologue for NVPs for XML input archives
+  template <class T> inline
+     void prologue(XMLInputArchive &, OptionalNameValuePair<T> const &)
+  { }
+
+  // ######################################################################
+  //! Epilogue for NVPs for XML output archives
+  /*! NVPs do not start or finish nodes - they just set up the names */
+  template <class T> inline
+     void epilogue(XMLOutputArchive &, OptionalNameValuePair<T> const &)
+  { }
+
+  //! Epilogue for NVPs for XML input archives
+  template <class T> inline
+     void epilogue(XMLInputArchive &, OptionalNameValuePair<T> const &)
+  { }
+
+  // ######################################################################
   //! Prologue for SizeTags for XML output archives
   /*! SizeTags do not start or finish nodes */
   template <class T> inline
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -65,6 +65,39 @@ namespace cereal
     return {name, std::forward<T>(value)};
   }
 
+  template <class T>
+  OptionalNameValuePair<T> make_optional_nvp(std::string const& name, T&& value)
+  {
+     static std::remove_reference_t<T> const defaultValue{};
+     return{name.c_str(), std::forward<T>(value), defaultValue};
+  }
+
+  template <class T, class V>
+  OptionalNameValuePair<T> make_optional_nvp(std::string const& name, T&& value,
+     V const& defaultValue)
+  {
+     return{name.c_str(), std::forward<T>(value), defaultValue};
+  }
+
+  template <class T>
+  OptionalNameValuePair<T> make_optional_nvp(char const* name, T&& value)
+  {
+     static std::remove_reference_t<T> const defaultValue{};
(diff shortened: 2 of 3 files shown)


sweeneychris:
+1

rcdailey:
I realize my code might be crappy, since I'm not super familiar with this code base, but I'm trying to proactively help add this feature. I'd appreciate some feedback from other contributors on how I can improve the code. I'd really like this functionality added upstream, because I have to patch this myself on my local copy of this library to get this functionality.

Feedback is very appreciated. Let me know if this approach works, and if so, if we can get it into the mainline.

mikke89:
I just discovered cereal and like it a lot so far. This was one of those features I quickly looked for, so thanks for the pull request, I hope this feature gets implemented soon.

Meanwhile I've used your code and implemented support for the JSON archive. You can find the patch here: https://gist.github.com/mikke89/6ae0e8a95a66e06b3b8a4fa4bcee8fcc
Use it as you like, it's probably not production ready, but works for me :)

mwoelk83:
I'm using this and mikke89's changes (thanks to both). But I had to change the loadvalue(..) in json.hpp functions too. 

from (example)

`      void loadValue(bool & val)        { search(); val = itsIteratorStack.back().value().GetBool(); ++itsIteratorStack.back(); }`
to

`      void loadValue(bool & val)        { search(); if (itsLoadOptional) return; val = itsIteratorStack.back().value().GetBool(); ++itsIteratorStack.back(); }`

Reason is that I got a crash when I have more optional values (not only one) that are not found in archive (so that they are loaded optionally). That's because `itsIndex` in `Iterator` is incremented (see `++itsIteratorStack.back()`) and than the check 

```
          auto const actualName = itsIteratorStack.back().name();

          // Do a search if we don't see a name coming up, or if the names don't match++		 
		  if (!actualName || std::strcmp(itsNextName, actualName) != 0) {
```

fails (invalid memory access).

Type1J:
Any updates here?

wolfwoolford:
Getting these modifications working (pull request 328 and the suggest patch) was actually quite difficult.  But anyway they do not generalise to non integral types - if you try to load something like this from an  archive

std::chronos::seconds s;
cereal::make_optional_nvp("s", s,  timing::chronos::seconds{1});

you get big headaches.

Instead I went with an approach based on a comment in https://github.com/USCiLab/cereal/issues/30 (thanks erichkeane). 

Disclaimer: I'm not sure of what corner cases exist and if there are things I am not considering. I'm just hacking.

namespace cereal
{
    template <class Archive, class T> inline
    void optional_nvp( Archive& archive, const char* name, T& value )
    {   
        try
        {
            archive(make_nvp(name, value));
        }
        catch(cereal::Exception&)
        {
            archive.setNextName(nullptr);
            // Loading a key that doesn't exist results in an exception
            // Since "Not Found" is a valid condition for us, we swallow
            // this exception and the archive will not load anything
        }
    }   

    template <class Archive, class T> inline
    void nvp( Archive& archive, const char* name, T& value )
    {
        archive(make_nvp(name, value));
    }
} // namespace cereal
                        

struct class_one
{
    int x1=1, x2=2;
    std::chrono::milliseconds ms1 = std::chrono::milliseconds{10};
    std::chrono::milliseconds ms2 = std::chrono::milliseconds{20};
    std::map<std::string, std::string> map1;
    std::map<std::string, std::string> map2;

    template<class Archive>
    void serialize(Archive & archive)
    {   
        cereal::nvp(archive, "x1", x1);
        cereal::optional_nvp(archive, "x2", x2);
        cereal::nvp(archive, "ms1", ms1);
        cereal::optional_nvp(archive, "ms2", ms2);
        cereal::nvp(archive, "map1", map1);
        cereal::optional_nvp(archive, "map2", map2);
    }   
};

Type1J:
I really hope this is resolved soon. I've been trying to get back to main-line Cereal for over a year, now.

rcdailey:
I think it will be important to get feedback and guidance from the library author(s). At this point, no attention has been given to my PR. I've stopped using cereal since this PR, since I cannot rely on these new feature additions. As such I will no longer be working on this. I will close for now, but if the author ever responds, he/she can reopen and we can discuss this again.

PR #335, opened 2016-08-25
Register of chainable relations fix

dlardi:
There is a case when we could find several different relations to write to 'unregisteredRelations' map with one 'otherBase'.

### DIFF
--- a/include/cereal/details/polymorphic_impl.hpp
+++ b/include/cereal/details/polymorphic_impl.hpp
@@ -232,7 +232,7 @@ namespace cereal
         }
 
         // Find all chainable unregistered relations
-        std::map<std::type_index, std::pair<std::type_index, std::vector<PolymorphicCaster const *>>> unregisteredRelations;
+        std::multimap<std::type_index, std::pair<std::type_index, std::vector<PolymorphicCaster const *>>> unregisteredRelations;
         {
           auto checkRelation = [](std::type_index const & baseInfo, std::type_index const & derivedInfo)
           {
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Can you provide pseudo code or real code for an example demonstrating this so it can be added to a test case as well?

dlardi:
Example code:
```
struct A
{
  virtual ~A() {}
  template <class Archive>
  void serialize( Archive & )
  {
  }
};

struct B : A
{
  template <class Archive>
  {
    ar( cereal::base_class<A>(this) );
  }
};

struct C : B
{
  template <class Archive>
  {
    ar( cereal::base_class<B>( this ) );
  }
};

struct D : C
{
  template <class Archive>
  {
    ar( cereal::base_class<C>( this ) );
  }
};

CEREAL_REGISTER_TYPE(B)
CEREAL_REGISTER_TYPE(D)

template <class IArchive, class OArchive>
void test_polymorphic_relations()
{
  std::shared_ptr<A> o_shared = std::make_shared<D>();
  std::ostringstream os;
  {
    OArchive oar(os);
    oar( o_shared);
  }
}
```

std::multimap instead of std::map for unregistered relations will save some iterations. To check this you could extend the above hierarchy to say G class.

AzothAmmo:
Made some small changes when I merged this. I don't think the multimap is necessary, since it seems it was used to catch multiple potential paths during each iteration, when we really only desire the shortest new path. So went back to a map, but now check to see if we already have a shorter path we haven't commit to the main map we can replace.

I think your do-while loop was what was catching most of the missed paths.

PR #337, opened 2016-08-31
cmake: drop -Werror from c++ flags

junghans:
Putting `-Werror` in by default is very bad. Nobody knows what warning future compilers will show!


junghans:
Same here, broken in v1.2.1 with gcc-6.1 on OpenSUSE:
<https://build.opensuse.org/package/show/devel:libraries:c_c++/cereal>

Alternatively, you could add an cmake option, e.g DONT_INJECT_WERROR.

AzothAmmo:
I think the best solution to this would be adding a cmake option.

junghans:
Done

AzothAmmo:
Merged into develop; I modified the logic so that the option isn't a negation (DONT_INJECT -> WITH_WERROR). Default is to use Werror, as you had.

Issue #341, opened 2016-09-07
Compile error in traits.hpp on OSX


astrodroid:
Hi,
 
I am using cereal to load game scenario/map data from file. So far this was working very well. But a couple of days ago, I had to move the classes that cereal serialises away from the game project in a separate static library that is used to calculate the objects position etc..  

However as soon as I reference the cereal headers within that static library I encounter a compile error on the traits.hpp file:
```cpp
    namespace detail
    {
      struct shared_from_this_wrapper
      {
        template <class U>
        static auto **check**( U const & t ) -> decltype( ::cereal::access::shared_from_this(t), std::true_type() );  _//error: Expected member name or ';' after declaration specifiers_

        static auto **check**( ... ) -> decltype( std::false_type() ); _//error: Expected member name or ';' after declaration specifiers_

        template <class U>
        static auto get( U const & t ) -> decltype( t.shared_from_this() );
      };
    }

    //! Determine if T or any base class of T has inherited from std::enable_shared_from_this
    template<class T>
    struct has_shared_from_this : decltype(detail::shared_from_this_wrapper::**check**(std::declval<T>())) _//error: Expected unqualified-id_
    { };
```

The problem is a bit confusing as this only happens when compiling for OSX and not for iOS. If I understood what is happening correctly, it is not that the static check that is failing, but the compiler appears to be calling a macro called check contained within "AssertMacros.h" :

```cpp
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/
MacOSX10.11.sdk/usr/include/AssertMacros.h:1291:28: note: expanded from macro 'check'
        #define check(assertion)  __Check(assertion)
```

At the same time the same piece of code that doesn't work within the staticlibrary works fine if called from the main game project, for both iOS and OSX. 

For now I have worked-around this issue by renaming the check methods within traits.hpp to "check_" which resolves the issue for me.

This is the full error log:
http://pastebin.com/6GPH6D50

AzothAmmo:
See #273 for the same issue. I previously listed this as a nofix but might reconsider if this is going to continue to affect people.

astrodroid:
Ok thanks. 

Yes, setting ```#define __ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES 0 ``` in the prefix folder  prevents the error from occurring.

Really odd naming choice for that C macro.

erichkeane:
[...]

Issue #273, opened 2016-04-10
Name conflict: user defined check() macro conflicts with cereal's method name


daniel-zhang:
I have an error check macro that looks like this:
```
    #define check(_expression)\
        (void)(!!(_expression) || (_wassert(_CRT_WIDE(#_expression), _CRT_WIDE(__FILE__), __LINE__), 0))
```

Compilation fails If this macro appears before the inclusion of ``cereal/access.hp``, as it conflicts with  cereal's ``detail::shared_from_this_wrapper::check``. 

The issue is trivial. But It would be nice if cereal does not occupy beloved names like ``check``. At least make it ``_cereal_check_``

AzothAmmo:
Similar issue to #104. Trivial issues can have trivial pull requests! Is the word `test` (I'm not a fan of leading underscores) safe to use? Is it the onus of a library to avoid common words that the preprocessor might replace? 

Here's everywhere lowercase check shows up in our code:

```
./cereal/archives/xml.hpp:          We check to see if the specified NVP matches what the next automatically loaded node is.  If they
./cereal/archives/json.hpp:          check of our node state and decide whether we are in an array or an object.
./cereal/archives/json.hpp:          check to see if an NVP has been provided (with setNextName) and if so, see if that name matches the actual
./cereal/archives/json_backup.hpp:          check of our node state and decide whether we are in an array or an object.
./cereal/archives/json_backup.hpp:          check to see if an NVP has been provided (with setNextName) and if so, see if that name matches the actual
./cereal/details/polymorphic_impl.hpp:        Since the compiler needs to check all possible overloads, the
./cereal/details/traits.hpp:      typedef typename detail::has_member_save_impl<T, A> check;
./cereal/details/traits.hpp:      static_assert( check::value || !check::not_const_type,
./cereal/details/traits.hpp:      typedef typename detail::has_member_versioned_save_impl<T, A> check;
./cereal/details/traits.hpp:      static_assert( check::value || !check::not_const_type,
./cereal/details/traits.hpp:      using check = typename detail::has_non_member_##test_name##_impl<T, A>;                                                \
./cereal/details/traits.hpp:      static_assert( check::value || !check::not_const_type,                                                                 \
./cereal/details/traits.hpp:      using check = typename detail::has_member_##test_name##_impl<T, A>;                                            \
./cereal/details/traits.hpp:      static_assert( check::valid,                                                                                   \
[...]

Issue #104, opened 2014-06-19
Getting compilation error in Xcode 5.1 (Clang based on LLVM 3.4)


valleyman86:
I am getting the following error during compilation after upgrading to 1.0.0 from 0.9.1 Cereal. I can't seem to figure out the cause but it feels like it may be associated with C++ 11. This same code worked before and all i did was update the library. I tried running the basic sample from the readme and that failed to load the cereal header as well. Any ideas?

AnimationSerializer/cereal/details/traits.hpp:1176:21: Expected member name or ';' after declaration specifiers

AnimationSerializer/AnimationSerializer/RTSAnimationSpriteData.mm:9:9: In file included from AnimationSerializer/AnimationSerializer/RTSAnimationSpriteData.mm:9:

AnimationSerializer/AnimationSerializer/RTSAnimationSpriteData.h:12:10: In file included from AnimationSerializer/AnimationSerializer/RTSAnimationSpriteData.h:12:

AnimationSerializer/AnimationSerializer/RTS/RTSAnimationSpriteData.h:13:10: In file included from AnimationSerializer/AnimationSerializer/RTS/RTSAnimationSpriteData.h:13:

AnimationSerializer/cereal/types/string.hpp:33:10: In file included from AnimationSerializer/cereal/types/string.hpp:33:

AnimationSerializer/cereal/cereal.hpp:42:10: In file included from AnimationSerializer/cereal/cereal.hpp:42:

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/usr/include/AssertMacros.h:1291:28: Expanded from macro 'check'

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/usr/include/AssertMacros.h:280:5: Expanded from macro '__Check'

pmoulon:
@AzothAmmo Any news about clang compatibility (since #75)?

AzothAmmo:
We don't have any compatibility issues that I am aware of other that the issue here which is fixed using the macro mentioned. #75 only refers to our automated build system, which we haven't taken the time to configure for clang.

winduptoy:
What's the current status of this?

I can't compile with Xcode/Clang either. I've done nothing besides importing cereal.

winduptoy:
[...]

PR #342, opened 2016-09-07
Avoid accidentally using the "check" macro on OSX

erichkeane:
Fix for #341, #273, and #104.  Replace usages of "check(" with
"(check)(", which will prevent us from colliding with the macro defined
in OSX's Assert macros.

### DIFF
--- a/include/cereal/details/traits.hpp
+++ b/include/cereal/details/traits.hpp
@@ -1180,9 +1180,9 @@ namespace cereal
       struct shared_from_this_wrapper
       {
         template <class U>
-        static auto check( U const & t ) -> decltype( ::cereal::access::shared_from_this(t), std::true_type() );
+        static auto (check)( U const & t ) -> decltype( ::cereal::access::shared_from_this(t), std::true_type() );
 
-        static auto check( ... ) -> decltype( std::false_type() );
+        static auto (check)( ... ) -> decltype( std::false_type() );
 
         template <class U>
         static auto get( U const & t ) -> decltype( t.shared_from_this() );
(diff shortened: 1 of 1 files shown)


erichkeane:
@astrodroid , if you coudl validate this on OSX, it woudl be greatly appreciated.

AzothAmmo:
Will merge as soon as someone on OSX can test.

astrodroid:
@erichkeane, I will be away for most of today, I can test this pull request this evening or tomorrow.

astrodroid:
Hi although it works fine for me, I haven't been able to run any unit tests as I haven't found a way to run all of them automatically. Is there a script that I missed or do I have to call each test case?

erichkeane:
There is a CMake file in the root directory, and the unit tests all build (and I think run?) when doing 'make' after that.

Do something like this from the root of Cereal (assuming you have cmake installed):
mkdir Build
cd Build
cmake ../
make

astrodroid:
Thanks @erichkeane, I have cmake installed. 

I did run cmake with the cmake_file.txt modified as here so that boost will be found: 

```
set(Boost_USE_STATIC_LIBS TRUE)
set (BOOST_ROOT /Users/andreas/Documents/Development/build/)
set(Boost_INCLUDE_DIR /Users/andreas/Documents/Development/boost)
set(Boost_LIBRARY_DIR /Users/andreas/Documents/Development/boost/build/lib)
```
but I get link time error, appearing something has gone wrong in the setting up of the environmental variables.

```
Linking CXX executable test_array
Undefined symbols for architecture x86_64:
  "boost::unit_test::unit_test_main(bool (*)(), int, char**)", referenced from:
      _main in array.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [unittests/test_array] Error 1
make[1]: *** [unittests/CMakeFiles/test_array.dir/all] Error 2
make: *** [all] Error 2
```

If some of you has

astrodroid:
OK I managed to run the tests. 

It passes all of them except the portability_test 32 and 64 bit which returns segmentation fault: 11. 

But that was present even before and I haven't tried to look up what's causing it.

AzothAmmo:
Merged into develop.

The portability tests are probably because you don't have a 32bit compiler installed, you can safely compile with those tests turned off (SKIP_PORTABILITY_TEST=ON) for CMAKE.

PR #345, opened 2016-09-16
Use statics initialization on demand fixed #331

### LINKED ISSUE #331: Crash if setting CEREAL_THREAD_SAFE=1 with polymorphic class
Run the polymorphic example (from: http://uscilab.github.io/cereal/polymorphism.html)
with `CEREAL_THREAD_SAFE=1`. #320 

The result crash on init at (polymorphic_impl.hpp line 413)
`auto lock = StaticObject<InputBindingMap<Archive>>::lock();`

auric:
Fixed crash CEREAL_THREAD_SAFE=1 with polymorphic class. Crash happened because StaticObject<T>::instanceMutex not initialized when StaticObject<T>::lock() called for the first time. Basically it is better to use static vars inside static functions to have predefined order of construction like it is done in StaticObject<T>::getInstance(), which calls create and forces instantiation at pre-execution time.
Also in this commit removed static T &instance; since it is possible that there were be two instances created
1) template <class T> T & StaticObject<T>::instance = StaticObject<T>::create();
2) inside StaticObject<T>::create()

### DIFF
--- a/include/cereal/details/static_object.hpp
+++ b/include/cereal/details/static_object.hpp
@@ -115,6 +115,7 @@ namespace cereal
           return LockGuard{};
           #endif
         }
+
       private:
         static T & instance;
     };
--- a/include/cereal/details/static_object.hpp
+++ b/include/cereal/details/static_object.hpp
@@ -120,7 +120,6 @@ namespace cereal
     };
 
     template <class T> T & StaticObject<T>::instance = StaticObject<T>::create();
-
   } // namespace detail
 } // namespace cereal
 
--- a/include/cereal/details/static_object.hpp
+++ b/include/cereal/details/static_object.hpp
@@ -109,7 +109,7 @@ namespace cereal
         static LockGuard lock()
         {
           #if CEREAL_THREAD_SAFE
-          std::mutex instanceMutex;
+          static std::mutex instanceMutex;
           return LockGuard{instanceMutex};
           #else
           return LockGuard{};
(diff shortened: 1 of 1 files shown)


AzothAmmo:
This is causing two tests to fail - test_structs_specialized and test_polymorphic. I'm not sure why the Travis build didn't fail. Can you look into this?

AzothAmmo:
Bumping this to a later milestone so it can be investigated further.

erichkeane:
The mutex in lock should still be static, otherwise you're giving a lockguard with a dangling reference.

I played around with it a bit, but don't get why the tests are failing.  There is a massive difference in behavior between doing instanciate on instance vs t.  Just doing 't' barely calls it at all, doing instance results in a number of cases where PolymorphicCasters get initialized, but I don't get that code enough to understand what is going on.

AzothAmmo:
I can go ahead with the mutex change. The instantiate(instance) is part of the black magic for using overload resolution to fill out the polymorphic metadata.

erichkeane:
Seems right.  Note that line 112 needs to be static std::mutex....  Current patch is missing the static.

auric:
Of course, I write static, while adding this patch. Btw, I'd like to say that there is some problems with this patch, which I cannot understand.
It should be correct, but there were some errors during deserialization/serialization, if 
`static T & instance;` removed. I do not why but everything works correct only if `create()` function unchanged (I mean `instance` member is not removed)
```
        static T & create()
        {
          static T t;
          instantiate(instance);
          return t;
        }
```
Since I cannot understand what is going on here, I propose to do not change `create()`. But it would be better if someone explain what is happening here :)

AzothAmmo:
I went ahead and merged this. I'll try and make a write-up for how the static objects are initialized sometime, it's been a while and I need to look over the code to remember how it all works.

PR #351, opened 2016-10-08
Replaced hardcoded rapidsjon namespacing with CEREAL_RAPIDJSON_NAMESPACE

Enhex:
I've ran into a problem in which I had another library that uses RapidJSON too, but an older version of it and that caused runtime errors.

I noticed Cereal uses CEREAL_RAPIDJSON_NAMESPACE macro to define RapidJSON's namepsace, but some places had the namespace "rapidjson" hardcoded, so I replaced them with the macro.

This way it's enough to only change the macro to put Cereal's RapidJSON in a different namespace.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -96,8 +96,8 @@ namespace cereal
   {
     enum class NodeType { StartObject, InObject, StartArray, InArray };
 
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Merged into develop, thanks for noticing this.

PR #355, opened 2016-10-28
Fix bitset loading when bits are already set.

drivehappy:
While investigating porting our codebase from Boost.Serialization, noticed this issue. When loading into a bitset that may already have bits set high, the de-serialization won't properly set any bits that are low.


erichkeane:
IMO, Instead of this solution, we'd likely be better off doing a std::bitset::reset() at the beginning.  It would likely be dozens of times faster.

erichkeane:
+1, LGTM.

PR #357, opened 2016-11-07
Introduce singleLine mode

cygnus9:
I need to serialize JSON to a single line and want to use cereal to do that. So I added support for that.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -97,7 +97,6 @@ namespace cereal
     enum class NodeType { StartObject, InObject, StartArray, InArray };
 
     using WriteStream = rapidjson::OStreamWrapper;
-    using JSONWriter = rapidjson::PrettyWriter<WriteStream>;
 
     public:
       /*! @name Common Functionality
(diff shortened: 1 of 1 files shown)


erichkeane:
I had introduced pr#146 in the past which solves a very similar problem in a very different way :) perhaps useful to a discussion.

cygnus9:
You're implementation is fine too. Stupid that I didn't check if anyone attempted this before. The reason I did it this  way is that I didn't want to create different type for the single line  minimal mode since there's already an Options class that you can use to tweak your JSONOutputArchive instance.
I also think it's lame (no offense) to have different types just because RapidJSON has two distinct types. My initial implementation actually relied on the fact that the RapidJSON PrettyWriter inherits from Writer, so I thought I could just get a Writer pointer to whatever instance I constructed and I'd magically get the correct behaviour. Unfortunately the Writer class has no virtual functions, so the whole PrettyWriter deriving from Writer could have been soved by composition rather than inheritance. But anyway, that's just the way it is.

Anyway, I really hope either implementation is approved because I don't want to manually do all the serialization in my project.

jeffreyscottgraham:
Did either of these get merged?

PR #365, opened 2016-11-25
Doxygen access from extern

CasparKielwein:
We are using cereal as our serialization library together with our libraries and would like to be able to have our doxygen documentation link to that of cereal.

To do that, a tagfile is needed, and finer control over the output directory is useful.


AzothAmmo:
Merged to develop!

Issue #360, opened 2016-11-20
Archive::{is_loading,is_saving}::value


headupinclouds:
Is the following {`is_loading`,`is_saving`} type addition compatible with the cereal library?  This would greatly improve compatibility with existing boost serialization code and is seemingly unobtrusive.

```
template<class ArchiveType, std::uint32_t Flags = 0>
  class OutputArchive : public detail::OutputArchiveBase
{
  public:
    typedef bool_<false> is_loading;  // <- proposal
    typedef bool_<true> is_saving;     // <- proposal
   // ...
};
```

This supports the following serialization pattern:

```
struct Bar
{
    template <typename Archive> void serialize(Archive &ar, const std::uint32_t version)
    {
        std::int16_t tmp = foo;
        if(Archive::is_loading::value)
        {
            ar & tmp;
            foo = tmp; // For illustration only
        }
        else
        {
            tmp = foo; // For illustration only
            ar & tmp;
        }
        ar & foo;
    }
    std::int32_t foo;
};
```

AzothAmmo:
You can actually already get this information by simply using `std::is_base_of` with the archive and either `cereal::detail::InputArchiveBase` or `cereal::detail::OutputArchiveBase`, though I suspect what you are after is identical syntax to boost.

The best place to add this to cereal would be in `cereal.hpp` to the `InputArchive` and `OutputArchive` classes, something along the lines of

```cpp
using is_loading = std::true_type;
using is_saving  = std::false_type;
```

as appropriate, defined within the doxygen block for the `Boost Transition Layer`

If you choose to do a pull request adding this to cereal, please do so against the current develop branch and also include appropriate doxygen comments for the new values.

headupinclouds:
Thanks for the quick feedback (and a great library).  I will put together a PR over the holiday.  Having this change upstream will greatly simplify a boost -> cereal transition.

AzothAmmo:
PR merged.

PR #366, opened 2016-11-26
Add is_loading and is_saving values to cereal::{Input,Output}Archive

headupinclouds:
This provides compatibility for asymmetric serialize routines when transitioning from boost serialization and associated archive types.

See: https://github.com/USCiLab/cereal/issues/360

Example:
```
    template <typename Archive>
    void serialize(Archive &ar, const std::uint32_t version)
    {
        std::int16_t tmp;
        if(Archive::is_loading::value)
        {
            ar & tmp; // load 16 bit type from archive
            m_foo = inflate<std::int32_t>(tmp); // convert to 32 bit for processing
        }
        else
        {
            tmp = deflate<std::in16_t>(m_foo); // convert to 16 bit for storage
            ar & tmp; // save 16 bit
        }
    }
```

### DIFF
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -630,7 +630,7 @@ namespace cereal
           from boost, you can check this value within a member or external serialize function
           (i.e., Archive::is_loading::value) to enable behavior specific to loading, until 
           you can transition to split save/load or save_minimal/load_minimal functions */
-.      using is_loading = std::true_type;
+      using is_loading = std::true_type;
 
       //! Indicates this archive is not intended for saving
       /*! This ensures compatibility with boost archive types.  If you are transitioning
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -255,6 +255,20 @@ namespace cereal
           a large project from Boost to cereal.  The preferred interface for cereal is using operator(). */
       //! @{
 
+      //! Indicates this archive is not intended for loading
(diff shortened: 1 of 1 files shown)


include/cereal/cereal.hpp
```diff
@@ -611,6 +625,20 @@ namespace cereal
           a large project from Boost to cereal.  The preferred interface for cereal is using operator(). */
       //! @{
 
+      //! Indicates this archive is intended for loading
+      /*! This ensures compatibility with boost archive types.  If you are transitioning
+          from boost, you can check this value within a member or external serialize function
+          (i.e., Archive::is_loading::value) to enable behavior specific to loading, until 
+          you can transition to split save/load or save_minimal/load_minimal functions */
+.      using is_loading = std::true_type;
```
erichkeane: Build failed due to the extra '.' at the beginning of this line.

erichkeane:
Also, the ability to use SFINAE is likely better than your example!  Something like:
    template <typename Archive, typename std::enable_if<Archive::is_loading::value,int> =0>
    void serialize(Archive &ar, const std::uint32_t version)

include/cereal/cereal.hpp
headupinclouds: Thanks.  Fixed.

headupinclouds:
> Also, the ability to use SFINAE is likely better than your example!

Nice.

Let me know if you need any follow up.

Issue #329, opened 2016-08-10
Update continuous integration testing configurations


AzothAmmo:
Travis CI should test against every configuration we care about:

(For linux):

g++-4.7
g++-4.8
g++-4.9
g++-5
g++-6

clang++-3.3 
(newer clangs)

Clang should be tested against libstdc++ and libc++.

OSX builds should test clang.

Currently we are only doing a single g++ automated test.

Also need to get this set up for MSVC as well (#126), which should test against MSVC 2013 and MSVC 2015.

furkanusta:
Hi, I've been working on Travis improvements and did some refactoring here are the list of compilers:

on Linux:
* g++-4.8
* g++-5
* g++-6
* g++-7
* clang++-3.5
* clang++-3.6
* clang++-3.7
* clang++-3.8
* clang++-3.9
* clang++-4.0
* clang++-5.0

on OSX:
* clang XCode 7.3
* clang XCode 8
* clang XCode 8.1
* clang XCode 8.2
* clang XCode 8.3
* clang XCode 9

Currently it fails on two occasions
1. gcc7.1: reported in #416 and I will fix it with #427 
2. clang++-5.0: because json.hpp overrides the default CEREAL_RAPIDJSON_ASSERT to throw instead of using just assert and that conflicts with nothrow specifier (you can see it in the error log) but that is an another issue. (I can open an issue related to this)

Travis page: https://travis-ci.org/furkanusta/cereal
.travis.yml file: https://github.com/furkanusta/cereal/blob/travis/.travis.yml

If you have issues with the .travis.yml I can modify or I can send a pull request and we can discuss it there.
Moreover, after merging this I can try valgrind or code coverage integration

AzothAmmo:
This looks really nice. It would be great to get some of the g++ builds (at least one) also building the 32 bit tests. You can see my futile attempts to get this working in the [develop travis.yml](https://github.com/USCiLab/cereal/blob/develop/.travis.yml), but I wasn't able to get around an error for c++config.h not being found. We also (for now) still need g++-4.7 in the build list since it has some quirks not present in later versions.

furkanusta:
Hi, I've managed to enable 32bit-g++ build. Although there are a few problems Travis can currently check
Linux-GCC 4.7, 4.8, 4.9, 5.0, 6.0, 7.0 (64-bit)
Linux-GCC g++-5 (32-bit) [More can be added easily]
[...]

PR #367, opened 2016-11-27
Travis configuration updates

tusharpm:
- add the argument `-j4` to make using parallel jobs to speed up the build
- switch to trusty container-based infrastructure for Linux systems
- add a job to test with clang compiler (skips portability test)
- add a job to test on macOS with Xcode8

Partially addresses #329. Travis runs the following tests now:
- gcc-4.8.4 on Ubuntu 14.04 (trusty)
- clang-3.5.0 on Ubuntu 14.04 (trusty)
- clang-800.0.38 (Xcode 8) on osx10.11 (El Capitan)

Using parallel jobs for make and containers leaves enough time to potentially run valgrind reversing the effect of 1364e5f.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,38 @@
 language: cpp
+dist: trusty
+sudo: false
+group: beta
+
+addons:
+  apt:
+    sources:
+      - 'ubuntu-toolchain-r-test'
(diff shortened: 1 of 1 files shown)


AzothAmmo:
Thanks for this!

Issue #368, opened 2016-11-29
Using static/dynamic_cast instead of C-like cast


ntimeu:
This is mostly an observation and a question, but would it be possible to use only static_cast and dynamic_cast instead of the C-like "(type)" casting in Cereal ? We personally compile with the "-Wold-style-cast" flag, which with the "-Werror" flag gives us multiple errors.

Is it a good idea to use this new cast-style or not ?

erichkeane:
Already done in develop branch as of yesterday

AzothAmmo:
Will be included in the next release.

PR #363, opened 2016-11-23
Don't use C style casts

laudrup:
Change C style casts to C++ static casts.

This makes the code compile with the -Wold-style-cast warning enabled

Cereal uses fairly clean modern C++ style, but unfortunately a few headers use C style casts meaning they cannot be used without disabling warnings about these casts, using #pragmas or similar.

This shouldn't introduce any functional changes.

PR #369, opened 2016-11-29
Using C++11 static_cast instead of C-like casts.

ntimeu:
This PR enable the use of the "-Wold-style-cast". Now it is possible to compile source code with strict options in the target project. References #368 .


erichkeane:
Check out #363, which was just committed to develop. This seems to be a subset of it

AzothAmmo:
Will be part of the next release, just waiting on performing MSVC testing.

PR #397, opened 2017-04-11
Add remove_reference to determine whether template argument to BinaryData is const

### LINKED ISSUE #396: cereal::binary_data() with lvalue
If you pass an lvalue pointer to a const type to `cereal::binary_data()` the code does not compile with:

> include\cereal/details/helpers.hpp(217): error C2440: 'initializing': cannot convert from 'const int *' to 'void *'

Here is  a short code to reproduce the issue:

```
#include <fstream>

#include <cereal/cereal.hpp>
#include <cereal/archives/binary.hpp>

struct MyVector
{
public:
    const int* data() const { return data_; }
private:
    int data_[3];
};

template<typename Archive>
void save(Archive& ar, const MyVector& v)
{
    auto local_data = v.data();
    ar(cereal::binary_data(local_data, 3)); // fails
    //ar(cereal::binary_data(v.data(), 3)); // works
}

int main(int argc, char* argv[])
{
    std::ofstream file("test.dat", std::fstream::binary);

    cereal::BinaryOutputArchive archive(file);

    MyVector v;
    archive(v);
}
```

The reason for this is that `cereal::binary_data()` uses a "universal reference" as first parameter which is forwarded as `const int*&` in the example above which is afterwards not handled correctly in `struct BinaryData`

albertziegenhagel:
fixes #396

### DIFF
--- a/include/cereal/details/helpers.hpp
+++ b/include/cereal/details/helpers.hpp
@@ -210,7 +210,7 @@ namespace cereal
   {
     //! Internally store the pointer as a void *, keeping const if created with
     //! a const pointer
-    using PT = typename std::conditional<std::is_const<typename std::remove_pointer<T>::type>::value,
+    using PT = typename std::conditional<std::is_const<typename std::remove_pointer<typename std::remove_reference<T>::type>::type>::value,
                                          const void *,
                                          void *>::type;
 


AzothAmmo:
Looks good!

Issue #395, opened 2017-04-11
XML Archive Tag Names


hoensr:
Hello!

So, suppose I have this DTO:

```
struct Foo
{
    std::vector<int> BarList;

    template<class Archive>
    void serialize(Archive & archive)
    {
        archive(
            CEREAL_NVP(BarList)
            );
    }
};
```

Cereal will serialize it to XML like this:

```
<?xml version="1.0" encoding="utf-8"?>
<cereal>
        <BarList size="dynamic">
                <value0>18</value0>
                <value1>20</value1>
                <value2>2</value2>
                <value3>3</value3>
                <value4>4</value4>
        </BarList>
</cereal>

```

There are three things I don't like about this XML:

1. The "cereal" root tag... This is solved by [PR 321](https://github.com/USCiLab/cereal/pull/321).
2. I'd like to have a different inner tag name than the funny "value0", "value1", ...
3. I don't need the `size="dynamic"` attribute. What was it good for, anyway?

Therefore, I have forked the repo and added some patches for these issues. Of course, the default behaviour is unchanged, but since I didn't want to add another boolean argument (don't you just *love* these `Options(true, true, false, true, false)` constructors??), I went for sort of a fluent interface.

Now with this code:

```
struct Foo
{
    std::vector<int> BarList;

    template<class Archive>
    void serialize(Archive & archive)
    {
        archive(
            CEREAL_NVP(BarList).withInnerName("Bar")
            );
    }
};

    {
        Foo foo{ { 18, 20, 2, 3, 4 } };
        cereal::XMLOutputArchive xmlwrite(
            std::cout, 
            cereal::XMLOutputArchive::Options(4, true, false, "MyFoo").NoSizeAttributes());
        foo.serialize(xmlwrite);
    }

```

My XML looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<MyFoo>
        <BarList>
                <Bar>18</Bar>
                <Bar>20</Bar>
                <Bar>2</Bar>
                <Bar>3</Bar>
                <Bar>4</Bar>
        </BarList>
</MyFoo>

```

So, my additions were:
1. I have added the `withInnerName` function to the `NameValuePair` to control the name of the inner XML tags.
2. I have added the `NoSizeAttribute` function to the `XMLOutputArchive::Options` class to get rid of the attribute.

[...]

PR #400, opened 2017-04-12
Added withInnerName to set the names of the inner XML tags for lists

hoensr:
See also [this thread](https://github.com/USCiLab/cereal/issues/395).
This PR is about changing the inner node names of XML for lists.


Benjamin-Tan:
May I know when will this PR be merged? This feature is very useful as it doesn't name the vector with value0, value1, value2 ...

PR #401, opened 2017-04-12
Add option to turn off the size=dynamic attributes of lists

hoensr:
See also [this thread](https://github.com/USCiLab/cereal/issues/395)
This PR is about an option for suppressing the `size="dynamic"` attributes of lists in XML.

### DIFF
--- a/include/cereal/archives/xml.hpp
+++ b/include/cereal/archives/xml.hpp
@@ -119,13 +119,22 @@ namespace cereal
                             bool outputType = false ) :
             itsPrecision( precision ),
             itsIndent( indent ),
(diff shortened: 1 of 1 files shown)


AzothAmmo:
I made a few changes to the interface when I merged this. You can now do one of two things for options:

Either specify all of the options using the constructor, e.g.:
```cpp
cereal::XMLOutputArchive::Options(true, false, ... );
```

or use the new "named parameter" interface:

```cpp
cereal::XMLOuputArchive::Options().indent(true).sizeAttributes(false);
```

I removed an existing static function that returned an Options with a specific set of parameters.

Issue #417, opened 2017-06-30
Deserializing ptr to const objects


pkrenz:
The following example example fails to compile
```cpp

#include <iostream>

#include <cereal/archives/xml.hpp>
#include <cereal/types/memory.hpp>

class MyClass {
       public:
	MyClass() : myint_(0) {}
	MyClass(int input) : myint_(input) {}

	int myint_;

	template <class Archive>
	void save(Archive& arch) const {
		arch(myint_);
	}

	// incidentally, the load_and_construct mechanism also does not work
	// with pointers to constant objects
	/*template < class Archive >
	static
	void
	load_and_construct( Archive & arch, cereal::construct< MyClass > &
	construct )
	{
	  int input;
	  arch( input );
	  construct( input );
	}*/

	template <class Archive>
	void load(Archive& arch) {
		arch(myint_);
	}
};

void serialize_myclass() {
	std::shared_ptr<MyClass const> m(
	    new MyClass(5));  // this "const" is fine

	std::stringstream ss;
	{
		cereal::XMLOutputArchive arc1(ss);
		arc1(m);
	}

	std::shared_ptr<MyClass const>
	    m2;  // comment out "const" on this line and the code will compile.
	{
		cereal::XMLInputArchive arc2(ss);
		arc2(m2);
	}

	std::cout << "M2: " << m2->myint_ << std::endl;
}

int main() { serialize_myclass(); }
```

and produces the following error: 

```
In file included from main.cpp:4:0:
cereal/include/cereal/types/memory.hpp: In instantiation of ‘typename std::enable_if<(! cereal::traits::has_load_and_construct<T, Archive>::value), void>::type cereal::load(Archive&, cereal::memory_detail::PtrWrapper<std::shared_ptr<_Tp2>&>&) [with Archive = cereal::XMLInputArchive; T = const MyClass; typename std::enable_if<(! cereal::traits::has_load_and_construct<T, Archive>::value), void>::type = void]’:
[...]

PR #418, opened 2017-06-30
fix behavior for deserializing sharedptr to const obj

pkrenz:
Fix for issue #417

### DIFF
--- a/include/cereal/types/memory.hpp
+++ b/include/cereal/types/memory.hpp
@@ -278,14 +278,13 @@ namespace cereal
   typename std::enable_if<traits::has_load_and_construct<T, Archive>::value, void>::type
   CEREAL_LOAD_FUNCTION_NAME( Archive & ar, memory_detail::PtrWrapper<std::shared_ptr<T> &> & wrapper )
   {
-    auto & ptr = wrapper.ptr;
-
     uint32_t id;
 
     ar( CEREAL_NVP_("id", id) );
 
     if( id & detail::msb_32bit )
     {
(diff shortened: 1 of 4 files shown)


AzothAmmo:
There are a few issues with your proposed changes, which look good overall.

1. The code needs to compile using C++11 - you are using a few C++14 specific features (e.g. `std::decay_t'`) which we can't support in the current version of cereal.
2. Ideally we want pull requests done against the `develop` branch, since there are often other changes staged there before being merged into a stable release (`master)`.
3. It would be helpful, but it is not required (I'll do it otherwise) if you could add or modify unit tests to reproduce #417.
4. I'm a bit concerned that the new code creates a temporary `shared_ptr` before assigning into the wrapper. Perhaps a move assignment would be better here? Previously we were able to create a reference to the actual pointer, but it seems this is not possible if we use a cast.

include/cereal/types/memory.hpp
```diff
@@ -325,20 +325,20 @@ namespace cereal
...
-    auto & ptr = wrapper.ptr;
-
     uint32_t id;
 
     ar( CEREAL_NVP_("id", id) );
 
     if( id & detail::msb_32bit )
     {
-      ptr.reset( detail::Construct<T, Archive>::load_andor_construct() );
+      auto  ptr = std::const_pointer_cast<std::decay_t<T>>(wrapper.ptr);
```
WSoptics: This line appears superfluous to me, @pkrenz ?

WSoptics: What I mean is that `ptr` is `reset` in the line right after, so there's no point in assigning it a value, is there?

PR #422, opened 2017-07-19
add package.xml

artivis:
Add a package.xml so that cereal can be found / compiled by tools such as [catkin-tools](https://catkin-tools.readthedocs.io/en/latest/).

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>cereal</name>
+  <version>1.1.2</version>
+  <description>A C++11 library for serialization</description>
+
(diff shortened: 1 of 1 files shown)


artivis:
I did not filled the two fields `maintainer` and `author`, if anyone has a suggestion ?

AzothAmmo:
How are those fields usually interpreted? Do they imply the maintainer of the code itself, or of the package.xml file?

artivis:
You usually fill the author tag with all authors, duplicating it. Here there might be too many, therefore a common mention would do it. As for the maintainer, it is the main maintainer of the code itself, not only the package.xml file.

PR #426, opened 2017-08-26
Remove conflicting defines for Windows.h

soroush:
* Some Windows SDK installations (like mine) may define `max' in
    `Windows.h', causing `std::numeric_limits::max()' cease to work.

    Details: https://stackoverflow.com/q/27442885/275221

Signed-off-by: Soroush Rabiei <soroush.rabiei@gmail.com>


AzothAmmo:
Will merge after I fix the existing compilation error I introduced with defer.

Issue #416, opened 2017-06-28
make install fails on gcc 7.1.1-64bit


furkanusta:
Hi, I've just learned about this library and decided to try. However, 'make install' failed in various occasions

First:
cereal/archives/json.hpp:211:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
and failed because CMake sets -Werror by default.

After disabling it, it couldn't find  <gnu/stubs-32.h> in cereal/unittests/memory.cpp:28
After installing 32-bit development libraries for glibc (I am on a 64-bit system), it passed that one and failed at

> Linking CXX executable test_memory_32: 
>  cannot find -lgcc

I suppose I need to recompile gcc with multilib support to fix this because I couldn't find gcc-multilib in openSuse repositories. It would be quite useful if cereal didn't require 32-bit libraries in an 64-bit system.

I've installed the library and skipped tests for now, but I thought it would be useful to share my experience.

If any more information is needed to reproduce, I can provide

erichkeane:
Are you attempting to compile 'master' or a different directory?  The 28th line of memory.cpp is "#include "memory.hpp", which includes "common.hpp", but a grep of the source base doesn't show a single reference to stubs-32.h.

As far as the fallthrough, it seems that we disable it in rapidjson/internal/regex.h (possibly not re-enabling it?), but a fresh compile of cereal seems to work for me.  However, it DOES seem that we still have the problem in the code, so perhaps the disable is working?

erichkeane:
Additionally, I note the disable in regex.h is only #ifdef __clang__.

furkanusta:
First of all I am sorry that I got the stacktrace backwards and it looks like gnu/stubs came from type_traits in libstdc++, but 32-bit libraries still were a problem. I am not sure if it is a compiler, cmake or library bug but it required 32-bit libstdc++ and 32-bit gcc. After installing both of those, now make works fine.

I'll update here after make and the tests finish

AzothAmmo:
Apparently the fallthrough warning is new in GCC 7 (https://dzone.com/articles/implicit-fallthrough-in-gcc-7). We actually do want the switch statement to fall through here - since the [[fallthrough]] attribute is c++17, it looks like the easiest fix is to add a `/* fall through */` comment before appropriate cases.

cereal doesn't actually require 32 bit to work, we just try to build unit tests for both 32 and 64 bit. You can disable the 32 bit tests with the CMAKE flag `SKIP_PORTABILITY_TEST`.

[...]

PR #427, opened 2017-08-28
Fix for #416, compilation warnings for implicit-falltrough on gcc7.1

furkanusta:
After checking external/rapidjson/internal/regex.h, I've decided to copy same pragmas instead of relying on the inclusion of regex.h because regex.h pops back the pragmas at the end of the file (although not for gcc), and thus it would be better to change these pragmas per-file basis. 

It passes all the tests with gcc7.1.1 on my machine and a simple test code generates no warnings on clang4.0

However whole project fails to build on clang4.0 giving an unrelated error. I will check it as well and send a pull request if I can.

And a suggestion: maybe we can define a macro like CEREAL_FALLTHROUGH that expands to C++17 counterpart if it is defined and to nothing otherwise.

### DIFF
--- a/include/cereal/details/polymorphic_impl.hpp
+++ b/include/cereal/details/polymorphic_impl.hpp
@@ -273,7 +273,7 @@ namespace cereal
               return {path.size(), path};
             }
             else
-              return {std::numeric_limits<size_t>::max(), {}};
+              return {(std::numeric_limits<size_t>::max)(), {}};
           };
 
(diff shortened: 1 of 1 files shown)


furkanusta:
Ignore this patch for now, as it seems that this one fixes the problem for gcc7.1 but gives an error for previous versions since implicit-fallthrough is not available on them. I'll fix it and send an another patch.

Issue #348, opened 2016-09-22
Coverity Scan


kklouzal:
https://scan.coverity.com/

> Coverity Scan is a service by which Coverity provides the results of analysis on open source coding projects to open source code developers that have registered their products with Coverity Scan.

> Some examples of defects and vulnerabilities found by Coverity Quality Advisor include: 
> resources leaks 
> dereferences of NULL pointers 
> incorrect usage of APIs 
> use of uninitialized data 
> memory corruptions 
> buffer overruns 
> control flow issues 
> error handling issues 
> incorrect expressions 
> concurrency issues 
> insecure data handling 
> unsafe use of signed values 
> use of resources that have been freed

Easily integrated into your Travis CI build flow.

Coverity suggests creating a separate "coverity_scan" branch to push commits to whenever you want to trigger analysis but the master branch seems applicable since it's only updated ever so often.

What do you think @AzothAmmo ?

AzothAmmo:
I'm all for improvements to our automated testing, it's a fairly easy target for people to make contributions for as well. Also see https://github.com/USCiLab/cereal/projects/2

kklouzal:
Awesome! That's what I like to hear! I would go ahead and set the whole thing up however that would require access to Travis CI and for me to make a Coverity account for Cereal, things that a contributor doesn't have access to, you know? Good to hear your on board though. Coverity has helped me fix a few bugs now that I might not ever fix on my own, not to mention the countless things it brings to my attention I wasn't even aware of.

AzothAmmo:
You can always play around with Travis CI on your own fork of cereal.

furkanusta:
Hi, I've been working on adding coverage testing and successfully integrated coveralls.io which shows the code coverage of unittests.
You can see it in here: https://coveralls.io/github/furkanusta/cereal
travis file is in here: https://github.com/furkanusta/cereal/blob/coveralls/.travis.yml
The important part is 'after_success' part in matrix

Moreover I've checked Coverity Scan as suggested by @kklouzal, however it currently says that they will review the project to check whether it satisfies their open source guidelines before allowing me to check their findings. It says, it might take 1-2 business days.

First of all I am not a lawyer and I don't think they will strictly follow their Licensing Policy, nonetheless I didn't like it: https://scan.coverity.com/policy
> You will not disclose to any third party any comparison of the results of operation of Our Service or software products with other services or products, except as expressly permitted by this Agreement
[...]

PR #431, opened 2017-09-03
Refactored Travis integration and added new targets

furkanusta:
List of enabled targets:
On Linux:
+ gcc {4.7, 4.8, 4.9, 5.0, 6.0, 7.0} (without compat tests)
+ gcc 5.0 with compat tests enabled
+ clang {3.5, 3.6, 3.7, 3.8, 3.9, 4.0} (with libsdtc++)

On OSX:
+ XCode {7.3, 8, 8.2, 8.3, 9.0}

Missin Ones:
+ XCode 8.1 (Doesn't have CMake)
+ Linux Clang with libc++ (missing cxxabi.h)
+ Linux Clang 5.0 (missing <type_traits>)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,93 +1,177 @@
 language: cpp
+os: linux
 dist: trusty
 sudo: false
 group: beta
 
 addons:
   apt:
-    sources:
-      - 'ubuntu-toolchain-r-test'
-      - 'boost-latest'
-    packages:
-      - 'g++-multilib'
-      - 'libboost-serialization-dev'
+    sources: &trusty_default_sources
+      - ubuntu-toolchain-r-test
+      - boost-latest
+    packages: &trusty_default_packages
+      - libboost-serialization-dev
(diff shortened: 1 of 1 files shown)


furkanusta:
Currently this build will fail because of g++-7 as discussed in issue #416, I have a working fix for it in PR #427 but I am not sure if it breaks things related to Clang, I will push it again after checking it on Travis.

I've disabled some other targets that was failing and explained the reasons in more detail in #329. If this passes the review and I can fix the missing targets I am planning on adding Coverage scan (#348) or Valgrind support in Travis.

kklouzal:
I urge you to go with Coverity as it has proven to much more robust and sophisticated than Valgrind. I look forward to seeing it's integration with Cereal!!

Keep up the amazing work 💯 👍

AzothAmmo:
Merging, 7.0 bug should be fixed via #432 but I'll make sure it completes the checks.

PR #432, opened 2017-09-04
Fix for #416, Implicit falltrough warning on GCC7.1 (Re-sending)

furkanusta:
Although, I've opened a PR in #427, it got closed and says that it is merged. However, it also shows that no files are changed and I cannot find the commit message that shows it got merged. Thus, I am re-sending this pull request, with the fix that I've mentioned in that PR.

This one also fixes the build problem submitted in PR #431.

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -68,7 +68,7 @@ namespace cereal
 #include <vector>
 #include <string>
 
-#if defined(__clang__) or (defined(__GNUC__) and __GNUC__ >= 7)
+#if defined(__clang__) || (defined(__GNUC__) && __GNUC__ >= 7)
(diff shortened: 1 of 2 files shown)


include/cereal/archives/json.hpp
```diff
@@ -68,6 +68,11 @@ namespace cereal
 #include <vector>
 #include <string>
 
+#if defined(__clang__) or (defined(__GNUC__) and __GNUC__ >= 7)
```
AzothAmmo: I think the `or` and `and` need to be changed to boolean operators (`||` and `&&`) to make MSVC happy

furkanusta:
Fixed, I didn't even thought that would cause a problem, thus didn't check Appveyor results.

PR #438, opened 2017-09-12
Added coveralls.io support

furkanusta:
Hi, I've added coveralls.io support. It will build from travis and update the site. All you need to do is open a coveralls.io account and link the project. I've assumed it would be under account USCiLab/cereal but if you open coveralls account under an another name please don't forget to change the link in README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,19 @@ addons:
 matrix:
   include:
 
+    # |---------- Coveralls.io ----------|
+    - os: linux
+      compiler: gcc
+      addons:
+        sources: ['boost-latest']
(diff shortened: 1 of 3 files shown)


AzothAmmo:
Just FYI I will merge this but want to keep the option for a local valgrind build and need to decide whether to keep our own local coverage results or just link to coveralls.

furkanusta:
Oh, sorry about the valgrind part since most of the block was related to code coverage, I've missed that there was a custom valgrind target there.

By the way, you can run Valgrind via CTest without adding an additional target by using:
> ctest -T memcheck

PR #448, opened 2017-10-27
Add serialization support for C++17 std::optional and std::variant

arximboldi:
The test and code has been written in similar style to that of the existing support for `boost_variant`, I hope this is fine. For the tests to run one shall pass `-DCMAKE_CXX_STANDARD=17` to `cmake`.

### DIFF
--- a/include/cereal/types/optional.hpp
+++ b/include/cereal/types/optional.hpp
@@ -1,0 +1,64 @@
+/*! \file optional.hpp
+    \brief Support for std::optional
+    \ingroup OtherTypes */
+/*
+  Copyright (c) 2017, Juan Pedro Bolivar Puente
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+      * Redistributions of source code must retain the above copyright
+        notice, this list of conditions and the following disclaimer.
+      * Redistributions in binary form must reproduce the above copyright
+        notice, this list of conditions and the following disclaimer in the
+        documentation and/or other materials provided with the distribution.
+      * Neither the name of cereal nor the
+        names of its contributors may be used to endorse or promote products
+        derived from this software without specific prior written permission.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL RANDOLPH VOORHIES OR SHANE GRANT BE LIABLE FOR ANY
+  DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#ifndef CEREAL_TYPES_STD_OPTIONAL_
+#define CEREAL_TYPES_STD_OPTIONAL_
+
+#include <cereal/cereal.hpp>
+#include <optional>
+
+namespace cereal {
+  template <class Archive, typename T> inline
+  void CEREAL_SAVE_FUNCTION_NAME(Archive& ar, const std::optional<T>& optional)
+  {
+    if(optional) {
+      ar(CEREAL_NVP_("nullopt", true));
+    } else {
+      ar(CEREAL_NVP_("nullopt", false),
+         CEREAL_NVP_("value", *optional));
+    }
+  }
+
+  template <class Archive, typename T> inline
+  void CEREAL_LOAD_FUNCTION_NAME(Archive& ar, std::optional<T>& optional)
+  {
+    bool nullopt;
+    ar(CEREAL_NVP_("nullopt", nullopt));
+
+    if (nullopt) {
+      optional = std::nullopt;
+    } else {
+      T value;
+      ar(CEREAL_NVP_("value", value));
+      optional = std::move(value);
+    }
+  }
+} // namespace cereal
+
+#endif // CEREAL_TYPES_STD_OPTIONAL_
--- a/include/cereal/types/variant.hpp
+++ b/include/cereal/types/variant.hpp
@@ -1,0 +1,104 @@
+/*! \file variant.hpp
+    \brief Support for std::variant
+    \ingroup OtherTypes */
+/*
+  Copyright (c) 2014, 2017, Randolph Voorhies, Shane Grant, Juan Pedro
+  Bolivar Puente. All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+      * Redistributions of source code must retain the above copyright
+        notice, this list of conditions and the following disclaimer.
+      * Redistributions in binary form must reproduce the above copyright
+        notice, this list of conditions and the following disclaimer in the
+        documentation and/or other materials provided with the distribution.
+      * Neither the name of cereal nor the
+        names of its contributors may be used to endorse or promote products
+        derived from this software without specific prior written permission.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL RANDOLPH VOORHIES OR SHANE GRANT BE LIABLE FOR ANY
+  DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#ifndef CEREAL_TYPES_STD_VARIANT_HPP_
+#define CEREAL_TYPES_STD_VARIANT_HPP_
+
+#include "cereal/cereal.hpp"
+#include <variant>
+
+namespace cereal
+{
+  namespace variant_detail
+  {
+    //! @internal
+    template <class Archive>
+    struct variant_save_visitor
+    {
+      variant_save_visitor(Archive & ar_) : ar(ar_) {}
+
+      template<class T>
+        void operator()(T const & value) const
+        {
+          ar( CEREAL_NVP_("data", value) );
+        }
+
+      Archive & ar;
+    };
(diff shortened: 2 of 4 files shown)


patrikhuber:
👍 would love to see this merged as soon as possible! :-)

arximboldi:
Thanks! I'd like to add something in the spirit of #256 for `std::variant` too but I'd like to know what the maintainers think about it on what the best approach would be...

AzothAmmo:
This looks good overall - I think some of the files have been misplaced in your commit but I can easily resolve this. The develop branch has also moved to a slightly different form for the unittests where the test and implementation are split amongst hpp/cpp files, but again I can do this easily myself as well.

Regarding #256, which I honestly must have missed, feel free to make a separate PR regarding that kind of change.

AzothAmmo:
I just tried to fix everything and merge this but it was taking a bit longer than I expected. Here are detailed changes that need to be made:

1) PR should be against develop branch
2) test case "optional.cpp" moved to unittests directory
3) both test cases split into header (tests) and source (instantiation of tests) files, see all other tests for an example
4) optional and variant are both in the STL, so they should have the doxygen group "STLSupport" and not "OtherTypes"
5) Missing doxygen comments for functions in optional.hpp

I would also prefer to add a `CEREAL_HAS_CPP17` macro to `cereal/macros.hpp` and use that in place of checking `__cplusplus` value each time.

I actually did all of these but then got stuck on the next issue:

When we compile with a compiler that does not support C++17, the unit tests that require C++17 should not even be generated. Otherwise what we see in the output is a blank executable that passes all tests, when in actuality there is no test to perform. So we need to add both a list of tests that require C++17 as well as check the C++ version in CMAKE. I got stuck on the latter half of this, so if anyone knows a good way to perform this check (across all operating systems) please let me know.

Also this is a good example of how even an innocuous looking PR can turn into a lot of work!

patrikhuber:
> check the C++ version in CMAKE. I got stuck on the latter half of this, so if anyone knows a good way to perform this check (across all operating systems) please let me know.

I think [pybind11](https://github.com/pybind/pybind11) may have a solution for that, I think they detect the available C++ standard (and also allow the user to override it IIRC). You could have a look at their repo and probably the `PYBIND11_CPP_STANDARD` variable.
Sorry I can't offer a solution directly, but that's where I'd look.

arximboldi:
Wow @AzothAmmo that was a fast and thorough, thanks a lot for fixing all my issues--I take note so next time I hope to be more conforming.

With regard to the open question: could you just use the `CMAKE_CXX_STANDARD` variable that is passed to CMake? In that case it would be under control of the developer to say what the expected feature set is. Then the unit tests could be organized in subfolders and all tests in the `std17` subfolder are skipped if `CMAKE_CXX_STANDARD < 17`?

There is also this, but this is for core language features, no standard library features: https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

A last technique is to use the traditional `try_compile` CMake macro. Basically you pass it a simple program that uses the feature you want to check and see if it works (e.g. a simple program using `std::optional<>`). Here is more info: https://cmake.org/cmake/help/v3.0/command/try_compile.html

What do you think?

patrikhuber:
@arximboldi The fine-grained `CMAKE_CXX_KNOWN_FEATURES` are being phased out according to the cmake developers and shouldn't be used anymore. CMake 3.8 introduced the "meta"-features `cxx_std_11`, `cxx_std_14` and `cxx_std_17` to replace the fine-grained ones (https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html).

But yes I think reading the value of `CMAKE_CXX_STANDARD`, much like pybind11, should suffice in this case and I don't think there's a need for something fine-grained as `try_compile`. Note though that CMake only started to properly support `CMAKE_CXX_STANDARD 17` for the VS generator in 3.10, but I don't think that's a problem, since people using `<optional>` on VS are very forward-looking and early-adopters anyway (e.g. it needs `/std:c++17`) and would be on a recent CMake version too.

AzothAmmo:
`CMAKE_CXX_STANDARD` looks like the way to go - I was trying to figure this out without it being specified explicitly to CMAKE, but there's really no point in that.

include/cereal/types/optional.hpp
```diff
@@ -0,0 +1,64 @@
...
+#define CEREAL_TYPES_STD_OPTIONAL_
+
+#include <cereal/cereal.hpp>
+#include <optional>
+
+namespace cereal {
+  template <class Archive, typename T> inline
+  void CEREAL_SAVE_FUNCTION_NAME(Archive& ar, const std::optional<T>& optional)
+  {
+    if(optional) {
```
patrikhuber: I think this is the wrong way round. Shouldn't it be `if(!optional)`? I.e. if the optional is empty, then we store nullopt true, otherwise, we serialise nullopt false plus the content. @arximboldi

arximboldi:
You are right!  I changed the name of the member when adapting the
`boost::optional` code without realizing I was negating its meaning.
Sorry about that!

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

patrikhuber:
Ok, cool!

Hmm, after making the change in both `CEREAL_SAVE_FUNCTION_NAME` and `CEREAL_LOAD_FUNCTION_NAME`, I'm still having some issues. The value in the optional seems to get stored to the file (I'm using `BinaryArchive`), but when loading the same archive, `ar(CEREAL_NVP_("nullopt", nullopt));` always returns false. The code is quite simple obviously but I haven't found out yet where it goes wrong.

patrikhuber:
Forget what I said above, there was a stupid mistake. I shouldn't have negated the check in `CEREAL_LOAD_FUNCTION_NAME`, because it checks the boolean that's read there, not the actual optional. Only the check in `CEREAL_SAVE_FUNCTION_NAME` needs to be negated.

All working now :-) 👍

patrikhuber:
Hi, any updates on completing/merging this? :-O Would love to see it in the official master or develop branch as soon as possible, along with my fix #450.

arximboldi:
Are you talking to me @patrikhuber, is there any follow up I should act on?

patrikhuber:
@arximboldi I'm not sure actually :-) Maybe @AzothAmmo could let us know whether there's anything we should do?

patrikhuber:
Hmm actually I seem to get some errors on gcc-7, in `cereal/types/variant.hpp:34`: https://travis-ci.org/patrikhuber/eos/jobs/304081001#L6855

It compiles fine on clang-5 and VS2017.4 though!

furkanusta:
Are you sure it compiles fine with clang-5? In the line you linked, process starts with

/usr/bin/clang++-5.0 -I/usr/include/opencv ....
but it uses libstdc++ instead of libc++

patrikhuber:
@furkanusta clang uses libstdc++ by default, so yes it compiles fine with `clang-5` in the default configuration.
I haven't bothered with testing libc++, as last time I tried (well over a year ago though!), OpenCV didn't compile with it, so it was irrelevant to me.

furkanusta:
I was referring to line 6855
> cd /home/travis/build/patrikhuber/eos/build/examples && /usr/bin/clang++-5.0   -I/usr/include/opencv -I/home/travis/build/patrikhuber/eos/include -I/home/travis/build/patrikhuber/eos/3rdparty/cereal/include -I/home/travis/build/patrikhuber/eos/3rdparty/eigen -I/home/travis/build/patrikhuber/eos/3rdparty/glm -I/home/travis/build/patrikhuber/eos/3rdparty/nanoflann/include -I/home/travis/build/patrikhuber/eos/3rdparty/eigen3-nnls/src -I/home/travis/build/patrikhuber/eos/3rdparty/toml11  -std=c++1z -o CMakeFiles/generate-obj.dir/generate-obj.cpp.o -c /home/travis/build/patrikhuber/eos/examples/generate-obj.cpp

It looks like the failing build is using clang++-5.0, not g++-7

Because there is already a bug report for this issue for clang 
https://bugs.llvm.org/show_bug.cgi?id=33222
and also in debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=877838

According to debian, it is a clang bug and not because of cereal, though I am not sure if workarounds exist

patrikhuber:
@furkanusta Oh, you're absolutely correct, apologies. I read it exactly the opposite way. It compiles fine on g++-7, but not on clang-5. And you're correct, this has absolutely nothing to do with this PR, so ignore my comments regarding that.
That's quite an annoying bug actually. I encountered a bug with variant + libstdc++ earlier in the year with clang-4 but it then disappeared when I switched to clang-5, and everything was fine. This will be quite annoying, it probably means it'll be ages before we get to use `std::variant` on AppleClang/XCode... :-(

AzothAmmo:
Going to look into wrapping this up tonight

-----Original Message-----

AzothAmmo:
Blocked by #423 which fixes some warnings/other things that new compilers will complain about. Almost have that fixed but not quite.

AzothAmmo:
#423 is done, working on this now.

AzothAmmo:
Merged into develop, still needs documentation and some changes to travis.

patrikhuber:
Awesome, thank you!

Btw I saw some `#if __cplusplus >= 201703L` somewhere in the code. I think MSVC, even in 2017 (15.5.2) still defines `__cplusplus` as `199711L`: https://developercommunity.visualstudio.com/content/problem/139261/msvc-incorrectly-defines-cplusplus.html, https://developercommunity.visualstudio.com/content/problem/120156/-cplusplus-macro-still-defined-as-pre-c11-value.html.
I think it should probably be something like `#if __cplusplus >= 201703L || (defined(_MSC_VER) && _MSC_VER >= 1911)` or something like that, or maybe using `_MSVC_LANG` or `_HAS_CXX17`: https://twitter.com/StephanTLavavej/status/839179366315646976

include/cereal/types/variant.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include "cereal/cereal.hpp"
+#include <variant>
+
+namespace cereal
+{
+  namespace variant_detail
+  {
+    //! @internal
+    template <class Archive>
+    struct variant_save_visitor
```
fiesh: This clashes with the type of the exact same name in `boost_variant.hpp`.

fiesh:
I think this commit should be refactored.  It is basically a lot of copy and paste code with some adaptations.  It even clashes with `boost_variant.hpp` because the namespace is again called `variant_detail`.

patrikhuber:
@fiesh I'm not sure what you mean yet. The type is `cereal::variant_detail::variant_save_visitor`, so it's in the `cereal::` namespace. Why would a type `boost::variant_detail::variant_save_visitor` (if that exists, as you say) clash with `cereal::variant_detail::variant_save_visitor`? Maybe you're using some `using` directives in your code which causes the compiler not being able to disambiguate?

WSoptics:
@patrikhuber I am saying that https://github.com/USCiLab/cereal/blob/develop/include/cereal/types/variant.hpp cannot be simultaneously included with https://github.com/USCiLab/cereal/blob/develop/include/cereal/types/boost_variant.hpp

This is because of type names clashing, because the namespace was not changed.  Of course this can be fixed easily, but it remains copy and paste code.

fiesh:
@patrikhuber I am saying that https://github.com/USCiLab/cereal/blob/develop/include/cereal/types/variant.hpp cannot be simultaneously included with https://github.com/USCiLab/cereal/blob/develop/include/cereal/types/boost_variant.hpp

This is because of type names clashing, because the namespace was not changed.  Of course this can be fixed easily, but it remains copy and paste code.

patrikhuber:
Ooh I got it, okay!

fiesh:
@patrikhuber That's a nice approach I think.  I'm not entirely sure about the advantages of this approach over setting tags though.

I do think thought that having the user put something in the `detail` namespace is a design flaw that shouldn't be replicated.  The exact purpose of `detail` is to hide the implementation from the user.

patrikhuber:
@fiesh What exactly do you mean with settings tags? Did I miss something? :-)

Yes I sort-of agree with you on your namespace `detail` comment, but the namespace would be easily changeable - I posted this more because of the general idea. (And the pybind11 team might have a reason for the namespace choice :-) ).

fiesh:
@patrikhuber Wasn't meant as criticism towards the general idea, just a comment.

What I'm trying to say is I don't necessarily see how this is superior to setting tags for user provided types, i.e. providing template specializations that register used defined types for some sort of treatment.  I think this is a more common means to achieve something quite similar, but maybe I misunderstand something.

AzothAmmo:
I like the idea of using the same code for serialization if possible, similar to what we're currently doing with `map` and `unordered_map` (see [here](https://github.com/USCiLab/cereal/blob/develop/include/cereal/types/concepts/pair_associative_container.hpp)).

It looks like that for `variant` types, we'd need two bits of information: the class, and then a way of querying the size of the variant (e.g. `std::variant_size` or `boost::mpl::size`). For `optional`, we just need the class and the equivalent of `std::nullopt` (e.g. `boost::none`).

We could put the implementation details in types/concepts/optional.hpp or whatever, and then users (of which cereal would be one) would just write along the lines of the pydb `type_caster`.

AzothAmmo:
Considering this done for now; the consolidation of code for different flavors of variant or optional can be done in a new issue or pull request.

jonathansharman:
The optional serializer doesn't seem to work with non-default-constructible types, if even if a `load_and_construct` function is defined for those types. I don't totally understand the mechanics of `load_and_construct`, but would it be possible to use it to allow serialization of these types?

masariello:
> The optional serializer doesn't seem to work with non-default-constructible types, if even if a `load_and_construct` function is defined for those types. I don't totally understand the mechanics of `load_and_construct`, but would it be possible to use it to allow serialization of these types?

Possibly a consequence of #506 ?

PR #470, opened 2018-02-04
Change StaticObject instance management to hopefully avoid UBSAN

erichkeane:
The static-object hack currently used requires passing an
uninitialized reference as a reference.  This fix uses a cast
to void to hopefully avoid that undefined behavior.

### DIFF
--- a/include/cereal/details/static_object.hpp
+++ b/include/cereal/details/static_object.hpp
@@ -67,13 +67,12 @@ namespace cereal
     class CEREAL_DLL_EXPORT StaticObject
     {
       private:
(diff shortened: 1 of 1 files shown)


volcoma:
Have you tested if this would get optimized by the compiler when optimizations are enabled?

AzothAmmo:
@volcoma The CI checks are done with optimization on; for Windows we run both a debug build and a release (optimized) build. If it's passing those everything should in theory be good, but I'll check this one out locally first to double check.

AzothAmmo:
Seems like this still works; if it blows up down the line we can easily revert the change.

Issue #473, opened 2018-02-16
std::variant serialization clashes with boost serialization


fiesh:
The serialization routines for `std::variant` currently commited to `develop` clash with the `boost_variant` serialization because the same namespace was chosen, see pull request #448

AzothAmmo:
For now I'm just changing the namespaces but a new issue/PR can be opened to try and consolidate the code as discussed in #448.

patrikhuber:
@AzothAmmo I have one comment regarding the variant serialisation, and I think it fits here, instead of opening a new issue.
`types/boost_variant.hpp` uses `int32_t` to serialise the `which` index. `types/variant.hpp` serialises this with `auto` (save) and `decltype(variant.index())` (load). I am not sure that's a good idea in `types/variant.hpp`. According to [cppreference](http://en.cppreference.com/w/cpp/utility/variant/index), the type of `index()` is `std::size_t`. So I think that means it could be a different type on different platforms - or even be different on the same platform for 32/64 bit compilation. That means an archive with an `std::variant` is not portable, which in my opinion I would even consider a bug. Should we not use `std::uint32_t` for this?
Second, I think it would be great if all variant types used the same type for storing the index. This would mean that we can store something with an `std::variant` into an archive on one platform, and then read this archive on a platform like macOS (which does not have `std::variant`) with something like `boost::variant` or `mpark::variant`. This is a highly desired feature (and indeed something I use this for).

What do you think?
I'm happy to prepare a PR if it helps.

AzothAmmo:
@patrikhuber Good catch, we definitely want to be using fixed sizes when possible. If you want to do the PR go ahead!

PR #478, opened 2018-03-15
Changed serialised index type of std::variant to std::int32_t

patrikhuber:
This makes saving & loading std::variant cross-platform.

### DIFF
--- a/include/cereal/types/variant.hpp
+++ b/include/cereal/types/variant.hpp
@@ -32,6 +32,7 @@
 
 #include "cereal/cereal.hpp"
 #include <variant>
+#include <cstdint>
 
 namespace cereal
 {
@@ -80,7 +81,7 @@ namespace cereal
   template <class Archive, typename VariantType1, typename... VariantTypes> inline
   void CEREAL_SAVE_FUNCTION_NAME( Archive & ar, std::variant<VariantType1, VariantTypes...> const & variant )
   {
-    auto index = variant.index();
+    std::uint32_t index = variant.index();
     ar( CEREAL_NVP_("index", index) );
     variant_detail::variant_save_visitor<Archive> visitor(ar);
     std::visit(visitor, variant);
@@ -92,7 +93,7 @@ namespace cereal
   {
     using variant_t = typename std::variant<VariantTypes...>;
 
-    decltype(variant.index()) index;
+    std::uint32_t index;
     ar( CEREAL_NVP_("index", index) );
     if(index >= std::variant_size_v<variant_t>)
       throw Exception("Invalid 'index' selector when deserializing std::variant");
--- a/include/cereal/types/variant.hpp
+++ b/include/cereal/types/variant.hpp
@@ -81,7 +81,7 @@ namespace cereal
   template <class Archive, typename VariantType1, typename... VariantTypes> inline
   void CEREAL_SAVE_FUNCTION_NAME( Archive & ar, std::variant<VariantType1, VariantTypes...> const & variant )
   {
-    std::int32_t index = variant.index();
+    std::int32_t index = static_cast<std::int32_t>(variant.index());
     ar( CEREAL_NVP_("index", index) );
     variant_detail::variant_save_visitor<Archive> visitor(ar);
     std::visit(visitor, variant);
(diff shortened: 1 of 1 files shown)


patrikhuber:
Here's the basic change, as discussed in #473.

A few notes/comments:

* The function and template argument (`int target` and `int N`) are both `int`s. I am not sure whether this may now result in a type-conversion/narrowing warning if `std::uint32_t` is not `int`. But that would probably be only on exotic platforms? And previously it might have generated a warning too because `std::size_t` is probably not `int` on many platforms.

* I can't see a unit test for variant, i.e. no `test_variant`. Shall I copy & paste `test_boost_variant` here and add that?

patrikhuber:
I think something is not quite right yet with a88e1c7dfd2f1f1323694932df8c314af8f1777e. I've got VS2017 15.6, I confirmed that `/std:c++17` is enabled for `test_cpp17_variant`, but my `_MSVC_LANG` is `201403L`, not `201703L`.
Also I think `test_cpp17_variant`, here: https://github.com/USCiLab/cereal/blob/develop/unittests/cpp17/variant.cpp#L32 should either test for `_MSVC_LANG` too or use `CEREAL_HAS_CPP17` for the test (and the same for the `.hpp` file).
But either of these will not help as `_MSVC_LANG` is `201403L` for me, even with `/std:c++17`.

patrikhuber:
Okay, it seems the IDE is giving me the wrong tooltip value for `_MSVC_LANG`. It is indeed `201703L`.
Now I am just still confused why `CEREAL_HAS_CPP17 ` is not set if I include macros.hpp in the variant test .cpp file.

I am also confused why there is a .cpp and a .hpp file in the cpp17 test folder. The .hpp file seems to be never included, and both files have similar contents. What am I missing?

I also seem to have a problem that none of the tests seem to compile anymore... `Severity	Code	Description	Project	File	Line	Suppression State
Error	C2220	warning treated as error - no 'object' file generated	test_vector	cereal\include\cereal\external\rapidjson\document.h	110	
`. I got one of them to run an hour ago, and didn't really change much since. Hmm.

patrikhuber:
Okay, I think this is ready to merge. But please have a look at my comments above (the ones that are not strike-through). First and foremost I think we need to replace more instances of `#if __cplusplus >= 201703L` with the added `_MSVC_LANG` check.
Second, I really don't understand why there is a .cpp and .hpp file to each test, but maybe that's irrelevant.
I also really do not understand why `CEREAL_HAS_CPP17` seems not get defined on my platform if I include `"cereal/macros.hpp"` in the unittest, even though if I `#pragma message("_MSVC_LANG=" BOOST_PP_STRINGIZE(_MSVC_LANG))` it shows as `201703L`.

AzothAmmo:
Do you think there should be some documentation somewhere about how to turn on the C++17 tests? Since it's only for building the tests, maybe it would be sufficient as a comment in the CMakeLists.

Regarding the template parameter, hard to say what the best thing to do here is. I think most likely it will be fine left as `int` - this can always be changed later if it is ever an issue; it shouldn't ever be a breaking change.

The .hpp/.cpp is an oversight on my part - the original PR had the tests in the .cpp but the preferred way is to have the test logic in the .hpp and instantiate it in the .cpp, I'll fix that up.

You are right on the checks for C++17, they all need to be updated to have the MSVC friendly check.

patrikhuber:
> Do you think there should be some documentation somewhere about how to turn on the C++17 tests? Since it's only for building the tests, maybe it would be sufficient as a comment in the CMakeLists.

Hmm yea, really not sure. Maybe if there was a `CEREAL_USE_CPP17` flag or something, it would be a bit more obvious. But adding such a flag just for people that run the tests (mainly library developers I guess) may be a bit overkill too. Maybe just leave it I guess. I stumbled upon it but it's not too hard to figure out. I mainly missed it I think because they were in a subdirectory, and I had already scrolled down looking for something like `test_*optional*` - so I was just a bit blind :)

Ok great, yep agree w.r.t. the template params!

I think if you could fix the .hpp/.cpp part that would be great.
I can update the checks to be MSVC friendly next week - I'll do that in a separate PR, so that it's not in the way of merging this one.

PR #480, opened 2018-03-15
move specialize into separate smaller header

WSoptics:
When implementing `serialize` for a class as free function, no include is normally necessary.
When implementing `serialize` for a class as a member function, it's normally sufficient to forward declare `cereal::access` and declare it a `friend`.

However, when there is a need for specialization, one needs the base template so it can be specialized.  This means including `cereal/access.hpp`, which isn't that lightweight.  So I separated `specialize` into a separate, minimalist header that can be included in an effort to give includes as small a footprint as possible.

### DIFF
--- a/include/cereal/access.hpp
+++ b/include/cereal/access.hpp
@@ -1,5 +1,5 @@
 /*! \file access.hpp
-    \brief Access control, default construction, and serialization disambiguation */
+    \brief Access control and default construction */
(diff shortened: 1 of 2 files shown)


AzothAmmo:
I have no issues with this change since it doesn't break anything. I'll have to check if there is any documentation to patch up before merging it.

PR #485, opened 2018-03-21
Added const to the member and constructor of binaryinputarchive

nbelakovski:
Can't see a reason why it would need to modify the stream. Tests all seem to pass.


nbelakovski:
@AzothAmmo Any interest?

AzothAmmo:
I'm not entirely sure on this because while we currently interface via `rdbuf`, which is const, `istream` has a lot of other member functions that are not const. This would also need to be applied to the other archives and not `BinaryInputArchive` in isolation.

PR #492, opened 2018-04-06
Added support for C++17 std::variant

zurrutik:


zurrutik:
Ok, good to know, thanks!

Issue #346, opened 2016-09-21
Cannot Properly Include Library


kklouzal:
Taken from the **Quick Start** page:
>>installed anywhere you can access header files

I've gone ahead and dropped the "cereal" folder into my projects directory (the one which has my .h and .cpp files) then tried to include the necessary headers
```
#include "cereal\types\string.hpp"
#include "cereal\archives\portable_binary.hpp"
```
And my project explodes:
```
Severity	Code	Description	Project	File	Line	Suppression State
Warning	D9025	overriding '/sdl-' with '/GS-'	PeerNet	D:\Programming Backups\PeerNet\PeerNet\cl	1	
Error (active)		expected a declaration	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	285	
Error (active)		expected a ';'	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	302	
Error (active)		expected a ';'	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	314	
Error (active)		'constexpr' is not valid here	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	359	
Error (active)		expected a ';'	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	360	
Error (active)		'constexpr' is not valid here	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	781	
Error (active)		expected a ';'	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	781	
Error (active)		inline specifier allowed on function declarations only	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	1047	
Error (active)		namespace "chrono" has no member "duration"	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	1047	
Error (active)		expected a ';'	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	1047	
Error (active)		expected a declaration	PeerNet	c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\chrono	1117	
[...]

PR #496, opened 2018-04-30
Use relative instead of absolute includes

jhasse:
This makes it a lot easier to use cereal by just dropping the include folder in your project without changing any compile flags, as one can use `#include "third_party/cereal/cereal.hpp"` for example. This was requested here: https://github.com/USCiLab/cereal/issues/346#issuecomment-251151016.

### DIFF
--- a/include/cereal/access.hpp
+++ b/include/cereal/access.hpp
@@ -34,8 +34,8 @@
 #include <cstdint>
 #include <functional>
 
-#include "cereal/macros.hpp"
-#include "cereal/details/helpers.hpp"
+#include "macros.hpp"
+#include "details/helpers.hpp"
 
 namespace cereal
 {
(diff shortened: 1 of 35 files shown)


kklouzal:
Cereal includes a lot of files and folders. Keeping them all organized inside a 'cereal' folder instead of mish-mashing them all together with your other project files seems like the best approach. I've had the changes discussed in issue #346 committed to the library already so a drag and drop implementation is already possible. Again I think keeping cereal organized in it's own folder is the best approach.

nightlark:
Yes, this change would be really useful!

The two external libraries (rapidJson and rapidXML) included in cereal get this right, it makes creating installers for multiple platforms so much easier. It changes things from a question of "do we bundle cereal in an installer and clobber any copy already installed on the system that may be older or newer than the copy we need", and instead makes so that we can use an external/cereal subfolder just like cereal does for rapidJson and rapidXML.

(Actual choice made was not clobbering existing system libraries, but making the includes just like this PR does.)

PR #500, opened 2018-05-18
grant constexpr access to cereal class version

WSoptics:
This patch allows accessing the class version set with the `CEREAL_CLASS_VERSION` macro in a `constexpr` manner.

The use case is as follows:
Suppose you have a function `f` that does something with a type `A` that has a cereal version set.  Now whenever `A` changes, you want to make sure that `f` is updated accordingly.  A nice way to enforce this is to simply `static_assert(cereal::Version<A>::value == 2, "A has been changed, please update f accordingly!");`.  However, there was no way as of yet to access the version number once it was set.

### DIFF
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -164,6 +164,8 @@ namespace cereal
     #define CEREAL_UNUSED_FUNCTION static void unused() { (void)version; }
   #endif
 
+  template <typename T> struct Version;
+
   // ######################################################################
   //! Defines a class version for some type
   /*! Versioning information is optional and adds some small amount of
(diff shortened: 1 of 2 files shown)


AzothAmmo:
MSVC 2013 doesn't support `constexpr` so this would either need to be modified to work under MSVC 2013 or prevented from being compiled in that state. In [other places](https://github.com/USCiLab/cereal/search?utf8=✓&q=constexpr&type=) we used a macro to get around this.

fiesh:
I added a commit that only uses `constexpr` when available, according to https://msdn.microsoft.com/en-us/library/hh567368.aspx and https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

It might make sense to use the macro defines in `macros.hpp` in `polymorphic.hpp` as well, in particular since the macro in `polymorphic.hpp` seems to never use `constexpr` with MSVC, no matter its version.

WSoptics:
The AppVeyor build fails, not being able to find `boost/variant.hpp`.  This doesn't seem to be related to my commits?!

WSoptics:
Maybe a better solution altogether is to solve it like this:
```
struct S {
static constexpr uint32_t version = 123;
};
CEREAL_CLASS_VERSION(S, S::version)
```
and then use S::version instead of using cereal::Version<S>::value`.

If so, then I guess this pull request should be closed.  Thoughts?

WSoptics:
Will close as discussed above.

Issue #508, opened 2018-06-27
Do polymorphic types registering from source still fail with recent Visual Studio compiler?


uentity:
Hi!

As many others I'm adapting an existing serialization subsystem in my project from Boost -> Cereal. A big plus for Cereal is that I can actually read sources compared to Boost where I just give up :-)

My main question is in the issue title. What's wrong with Windows DLLs and/or Visual Studio that makes registering from sources impossible? Where is the source of the problem? It's not documented anywhere. Maybe it's already fixed in recent VS that has full C++14 support? Registering from sources is a HUGE advantage.

I was digging through the code and it seems that `CEREAL_BIND_TO_ARCHIVES(...)` is the trigger of pointer serialization code instantiation. If I place this macro in header (default registering from header approach) and then include this header in multiple cpp files, does it mean that serialization code will be instantiated multiple times in each translation unit?

I understand that static bindings map will store only one instance, and even "registering" should be called only once, because compiler will preserve only one instance of `bind_to_archives<...> const &` static variable. But what if I define serialization in one shared library and use it from another library (or executable) by including the header with `CEREAL_BIND_TO_ARCHIVES(...)`? Will it result in producing ptr serialization code in each library and multiple 'trials' to register it?

uentity:
Seems that devs aren't so active here...
While waiting for the answer I did check by myself. Polymorphic types registering in cpp-file that is built into shared library (DLL), that in turn is utilized by executable test-case app result in NO errors. Serialization working fine. Will further test how will two DLLs interact, one exporting serialization API and another using it.
Visual Studio 2017 15.7.2.

uentity:
@AzothAmmo thanks for the response!

BTW I've had to fix one minor issue for gcc/clang -- it's related to `CEREAL_DLL_EXPORT` expanding to nothing for these compilers. I build all modules in my project with `-fvisibility=hidden` which prevents all non-decorated symbols from being exported (like VS does by default). In that case you have to explicitly `#define CEREAL_DLL_EXPORT __attribute__ ((visibility("default")))`.

uentity:
Update: I've implemented 2nd approach and it's actually working just fine.

PR #521, opened 2018-09-25
Remove template parameter `Archive` from `InputBindingsMap` and `OutputBindingsMap`

uentity:
`Archive` isn't directly needed for any content in binding
structs and required only to instantiate different global static
maps for different archives.

This commit turns `InputBindingsMap` and `OutputBindingsMap` into
non-templated structs and modifies binding maps to include archive's
typeid, such that they describe relation: `typeid(Archive) ->
per-archive binding map`.

That way only two global static objects will be created in runtime (for
input and output bindings respectively) in each library/executable. Also
types of these globals are independent from included archives and known
beforehand.

### DIFF
--- a/include/cereal/details/polymorphic_impl.hpp
+++ b/include/cereal/details/polymorphic_impl.hpp
@@ -449,7 +449,6 @@ namespace cereal
         type, containing entries for every registered type that describe how to
         properly cast the type to its real type in polymorphic scenarios for
         shared_ptr, weak_ptr, and unique_ptr. */
-    template <class Archive>
     struct OutputBindingMap
     {
       //! A serializer function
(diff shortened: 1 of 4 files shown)


uentity:
Reason for this change is discussed in issue #508

uentity:
I know that there's one non-related commit with suppressing warnings, but what do you think about the main idea?

It literally changes nothing in public API, but at the same time allows me to unify serialization code across multiple shared libraries.

dtmoodie:
If this works as you say, it would be a great addition since I currently am doing some funky work arounds to make this work correctly across shared library boundaries.

uentity:
@dtmoodie I successfully apply this patch and use it in my project exactly to "unite" serialization code from multiple shared libraries at runtime.
Works fine. Because patch is relatively simple and non-affecting. Try it.

BotellaA:
Great PR!! @AzothAmmo What is missing to accept the merge?

PR #527, opened 2018-10-09
Gcc8 warnings fixes

WSoptics:
This keeps gcc8 from suggesting that `base_class_id_hash::operator()` be marked `noexcept` as well as from remarking that the case from type `char *` to `header *` increases the alignment requirements.  (The latter is per se a correct warning, but rapidxml makes sure the alignment is correct by hand, which the compiler can't know.)


WSoptics:
(The check failing is also present in the develop branch and not related to the merge request: https://ci.appveyor.com/project/AzothAmmo/cereal/branch/develop/job/61eo08pu3rm87ej7/messages)

PR #575, opened 2019-06-04
Fix fallthrough warning for JSON archive code.

RadixSeven:
I was getting a fallthrough warning on g++ while compiling the unit tests, so I added a macro that suppresses the warning. It should also do so for c++17.


AzothAmmo:
Previously fixed by #587

PR #579, opened 2019-07-24
Fixing various compilation warnings

AzothAmmo:
Updates rapidjson and doctest, so far.

Should hopefully fix many of the warnings as errors with newer compilers.


AzothAmmo:
Note that I'll be pushing ahead with the next release even though not all of the tests are passing - these will be addressed as soon as possible.

PR #587, opened 2019-10-22
Add fall through comments to json.hpp

stashlukj:
The -Werror flag is on by default in the CMakeLists.txt file. This will not compile when the project is included via CMake ExternalProject_Add with default settings.


AzothAmmo:
Essentially a dup of #416 and #432, but a bit nicer than turning off the warning for the file.

PR #595, opened 2019-11-06
supported clang-cl

durswd:
This PR makes tests and compile passed on clang-cl (clang on Windows)

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -286,7 +286,7 @@ namespace cereal
       void saveLong(T lu){ saveValue( static_cast<std::uint64_t>( lu ) ); }
 
     public:
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
       //! MSVC only long overload to current node
       void saveValue( unsigned long lu ){ saveLong( lu ); };
 #else // _MSC_VER
--- a/include/cereal/details/static_object.hpp
+++ b/include/cereal/details/static_object.hpp
@@ -44,7 +44,7 @@
     License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
     http://www.boost.org/LICENSE_1_0.txt) */
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
 #   define CEREAL_DLL_EXPORT __declspec(dllexport)
 #   define CEREAL_USED
 #else // clang or gcc
--- a/include/cereal/details/polymorphic_impl.hpp
+++ b/include/cereal/details/polymorphic_impl.hpp
@@ -654,7 +654,7 @@ namespace cereal
 
             auto ptr = PolymorphicCasters::template downcast<T>( dptr, baseInfo );
 
-            #ifdef _MSC_VER
+            #if defined(_MSC_VER) && !defined(__clang__)
             savePolymorphicSharedPtr( ar, ptr, ::cereal::traits::has_shared_from_this<T>::type() ); // MSVC doesn't like typename here
             #else // not _MSC_VER
             savePolymorphicSharedPtr( ar, ptr, typename ::cereal::traits::has_shared_from_this<T>::type() );


include/cereal/archives/json.hpp
```diff
@@ -286,7 +286,7 @@ namespace cereal
       void saveLong(T lu){ saveValue( static_cast<std::uint64_t>( lu ) ); }
 
     public:
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
```
erichkeane: I'm not sure you'd want to do this, clang-cl has the same integer sizes as MSVC, so we'd want the unsigned long overload as well here, right?

include/cereal/details/static_object.hpp
```diff
@@ -44,7 +44,7 @@
     License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
     http://www.boost.org/LICENSE_1_0.txt) */
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
```
erichkeane: clang-cl supports declspecs, so I'd imagine this is useful as well.

durswd: declspec causes this error on clang9

cereal::detail::StaticObject<cereal::detail::bind_to_archives<PolyDerivedLA, cereal::detail::(anonymous namespace)::polymorphic_binding_tag> >' must have external linkage when declared 'dllexport'

durswd: this error can be reproduced on test_polymorphic

include/cereal/archives/json.hpp
durswd: to suppress this error

```
call to member function 'saveValue' is ambiguous
```

in json.hpp

jdonald:
I was about to file a pull request then found this. We came up with byte-for-byte identical changes so fully approve here. @fnabulsi @bmcdb

We also were running an older version that lacked the `defined(__clang__) && defined(_MSC_VER)` case around line 560 of rapidjson.h, but what's in trunk is sufficient there now, so apparently rapidjson was ahead on clang-cl support while cereal just needs these three additional fixes.

include/cereal/archives/json.hpp
jdonald: Ok correction: in our org we never came across this even though our `json.hpp` looks the same. We're using clang-cl version 8.0.0. I wonder if @durswd has a different configuration whereby `sizeof(long)` is coming out as 8?

jdonald:
For some reason it doesn't let me add an inline comment far outside of the current changes, but on line 719 of polymorphic_impl.hpp we also had to change this:
```
      #if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
```
add add  `&& !defined(__clang__)`, else we run into errors with CEREAL_BIND_TO_ARCHIVES(). I suspect the reason we get different results compared to @durswd is a combination of issues: using a different version of clang-cl, have an older fork of of cereal, and we use different features of the library which affects which templates get exercised.

You can take a look at our comparable changes here: https://github.com/USCiLab/cereal/compare/master...jdonald:clang-cl

include/cereal/details/static_object.hpp
jdonald: We (myself + @fnabulsi + @bmcdb) got the identical error and had to make the same change, although I acknowledge this may warrant more investigation.

We are testing cereal only with static linking so it wouldn't surprise me if there's more work to ensure that clang-cl can properly build Cereal as a DLL.

AzothAmmo:
So what's the verdict here? Is this PR sufficient for supporting clang? Apparently we can build against [clang on appveyor](https://www.appveyor.com/docs/lang/cpp/), if anyone wants to figure that out.

durswd:
saveLong error causes when it is compiled as 64bit. it is no problem  when it is compiled as 32bit.

Instead of it

this changes uppresses this error
```
//! MSVC only long overload to current node
void saveValue( unsigned long lu ){ saveLong( lu ); };
```

->

```
//! MSVC only long overload to current node
void saveValue( unsigned long lu ){ saveLong( lu ); };
void saveValue( long lu ) { saveLong(lu); };
```

InBetweenNames:
The main issue with DLL support his that a tag in an anonymous namespace `polymorphic_binding_tag` is used as the template argument for `StaticObject`, which means internal linkage.  So, that it works on MSVC at all is actually a linker bug -- unless MSVC treats all anonymous namespaces as actually having the same "name" internally in the compiler, which would permit exporting.

InBetweenNames:
On thinking about this more: is it actually necessary to maintain the StaticObject as a `dllexport`?

InBetweenNames:
OK -- I found a solution for this.  It requires C++17, but essentially the following changes are made:

* `polymorphic_type_impl` no longer resides in an anonymous namespace
* `CEREAL_BIND_TO_ARCHIVES` now uses a `static inline` variable:

~~~

#define CEREAL_BIND_TO_ARCHIVES(...)                                     \
    namespace cereal {                                                   \
    namespace detail {                                                   \
    template<>                                                           \
    struct init_binding<__VA_ARGS__> {                                   \
        static inline bind_to_archives<__VA_ARGS__> const & b=           \
        ::cereal::detail::StaticObject<                                  \
            bind_to_archives<__VA_ARGS__>                                \
        >::getInstance().bind();                                         \
        CEREAL_BIND_TO_ARCHIVES_UNUSED_FUNCTION                          \
    };                                                                   \
    }} /* end namespaces */
~~~

The `StaticObject` template is still marked `dllexport`.

This enables the following:
* Each TU only accesses ONE `StaticObject` across the entire DLL
* Each DLL receives it's own StaticObject, but these are identical with correct usage of the Cereal macros (as documented)

I didn't realize this, but the reason `dllexport` is even used is to prevent the linker from optimizing away the static object.
It's not actually used for linkage purposes.

This change also reduces code size, as the `inline static` merges the definition across all TUs.

InBetweenNames:
The same approach could be used to get the `CEREAL_CLASS_VERSION` macro working on MSVC 2017, where right now you can get a multiple definition error.  That macro could be changed as follows (again requiring C++17)

~~~
  #define CEREAL_CLASS_VERSION(TYPE, VERSION_NUMBER)                                \
  namespace cereal { namespace detail {                                             \
    template <> struct Version<TYPE>                                                \
    {                                                                               \
      static inline const std::uint32_t version = Version<TYPE>::registerVersion(); \                                \
      static std::uint32_t registerVersion()                                        \
      {                                                                             \
        ::cereal::detail::StaticObject<Versions>::getInstance().mapping.emplace(    \
             std::type_index(typeid(TYPE)).hash_code(), VERSION_NUMBER );           \
        return VERSION_NUMBER;                                                      \
      }                                                                             \
      CEREAL_UNUSED_FUNCTION                                                        \
    }; /* end Version */                                                            \
                                                                                    \
~~~

InBetweenNames:
I have a proposed patch to fix this and a few other small linker-related annoyances on Windows in #657.

Issue #308, opened 2016-06-28
Avoid pretty-printing of JSON objects


rix0rrr:
I'd like to use Cereal to output JSON data in newline-separated record format (of course this semi-standard convention has a webpage: http://jsonlines.org/).

Right now, Cereal always outputs newlines while pretty-printing the JSON object, with no way to switch them off. This messes up my format.

What's the best way to get rid of the pretty-printing?

erichkeane:
At the moment, no real way to do so, but take a look at this PR https://github.com/USCiLab/cereal/pull/146 for how to do it.

m7thon:
You can simply replace `rapidjson::PrettyWriter` by `rapidjson::Writer` in the file cereal/archives/json.hpp. You then also need to comment out the call to `itsWriter.SetIndent`, since `rapidjson::Writer` has no `SetIndent`. That's all.

NuLL3rr0r:
I'm in favor of this feature, too. It would be nice if cereal provides a way for my application to dump the minified JSON in the logs and show pretty JSON on the UI.

petasis:
I also want this feature. Is there a way to avoid the unnecessary new lines in the outpout?

sebeaumont:
This would be very handy... why not add another option for minimised json?

drodgers-ls:
Seriously, this needs to be a configurable option, without needing to modify source code.  Really only need two parameters, indent size (int, default 4) and end of line (string, default "\n").  Or just provide a single on/off.  I'm using this library for a messaging protocol in an embedded system with constrained resources, and every byte counts.

jeffreyscottgraham:
I agree we need an easy obvious way to avoid pretty printing: Consider making pretty printing special and non-formatting the default!

sebeaumont:
Alors? Milestone?

avibahra:
Similar vain, need the most compact json for client/server communication.

erichkeane:
FWIW, I had a patch here: https://github.com/USCiLab/cereal/pull/146
that provided for the ability to disable pretty-printing for JSON.  It got held up at the time because @AzothAmmo was evaluating a newer RapidJSON and then I stopped using Cereal for anything production so it got dropped by me as well.

Someone could likely rebase that and get this functionality.

jeffreyscottgraham:
Has this been implemented yet? If so how to generate this compact output?

drodgers-ls:
> Has this been implemented yet? If so how to generate this compact output?

[...]

PR #613, opened 2020-01-30
PrettyJSONOutputArchive + MinimalJSONOutputArchive + PGI Compiler fixes

jrmadsen:
- `JSONOutputArchive` is aliased to `PrettyJSONOutputArchive` for backwards-compatibility
  - See #308 for full explanation of changes
- a minor addition to CMake that suppresses some warnings when used as subproject
- PGI compiler does not support `__extension__ typedef unsigned __int128 uint128` so excluded that from pre-processor


jeffreyscottgraham:
Merge please

jrmadsen:
Ah, I pushed the last commit for my local submodule forgetting this existed. I can remove it but it is pretty harmless, it just makes it such that you can provide the archive type via a type-trait. Not that I have any hope of actually getting a response to this PR at this point. 

@jeffreyscottgraham the admins seem particularly uninvolved so I would recommend that if is possible to just fork this repo, then do a `git remote add json-fixes` and `git pull json-fixes timemory` and then you can just make your fork a git submodule for whatever project you are working on and just install it alongside your package (if needed). It's not an ideal solution but probably the best when packages are no longer actively being developed. Which, based on their releases, it appears to be the case for cereal.

jeffreyscottgraham:
Yea, not impressed. Replaced all my cereal code with nlohmann's json lib.

Issue #638, opened 2020-04-22
Cereal and C++20


lukaszgemborowski:
I'm trying to use cereal library in my C++20 pet project (using current gcc 10 snapshot). As far as the library is considered there's only one issue (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html), implicit 'this' capture in two lambdas in cereal.hpp which results in a warning. Quite easy to fix. But there are two more "complicated" issues within tests:

1. basic_string.cpp - as far as I understand the error is in doctest `has_insertion_operator` trait. The root cause of this is: `operator<<(basic_ostream<char, _Traits>&, wchar_t) = delete;` (yes, this is explicitly deleted in C++20 stdlib). Assuming we aim at C++11 and onwards we can replace the check with slightly modified version of answer from SO: https://stackoverflow.com/a/22759544 - at least it won't fail to compile with C++2a enabled. The question is if we want to modify it in place or rather target to fix it upstream?

2. performance.cpp - this one I don't have idea how to approach. The problem is in boost library, basically boost::format won't compile with C++20 because it use two argument Allocator::allocate somewhere deep. This was removed in C++20.

Any thoughts?

lukaszgemborowski:
doctest fix is now merged into the dev branch. In parallel I was trying to use boost::format() from the latest boost develop branch and it seem to compile well with -std=c++20 switch on gcc 10. I'll try to setup a fresh environment to compile cereal with latest versions of boost and doctest and see if everything's fine, potentially closing this issue.

AzothAmmo:
Great – thanks for the update. If this looks to fix things I’ll update our internal doctest. From: Łukasz GemborowskiSent: Thursday, June 11, 2020 9:30 AMTo: USCiLab/cerealCc: SubscribedSubject: Re: [USCiLab/cereal] Cereal and C++20 (#638) doctest fix is now merged into the dev branch. In parallel I was trying to use boost::format() from the latest boost develop branch and it seem to compile well with -std=c++20 switch on gcc 10. I'll try to setup a fresh environment to compile cereal with latest versions of boost and doctest and see if everything's fine, potentially closing this issue.—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #639, opened 2020-04-22
Cleanly compile cereal for C++20

lukaszgemborowski:
This is possible solution to problem outlined in issue #638 
Tested on `g++-10 (Ubuntu 10-20200416-0ubuntu1~18.04) 10.0.1 20200416 (experimental) [master revision 3c3f12e2a76:dcee354ce56:44b326839d864fc10c459916abcc97f35a9ac3de]` , `c++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0` and `g++ (GCC) 9.3.0` with `cmake -DSKIP_PORTABILITY_TEST=ON`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,10 @@ else()
 
 endif()
 
+if (CMAKE_CXX_STANDARD GREATER 17)
+    set(SKIP_PERFORMANCE_COMPARISON ON)
+endif()
+
 if(NOT CMAKE_VERSION VERSION_LESS 3.0)
     add_library(cereal INTERFACE)
     target_include_directories(cereal INTERFACE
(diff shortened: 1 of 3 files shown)


AzothAmmo:
Regarding the doctest changes - has this been fixed in upstream doctest? That'd be the best way to address this instead of us having to maintain changes to our local embedded doctest.

Also, with regards to boost, this again sounds like something that is/will be fixed in a later release of boost, especially in regards to removed C++20 features.

I think it'd be best if you could break out your changes to two PRs- one with the fix for the lambda capture and one with the other changes. I'll merge the lambda change asap, the others I'd rather delay until checking in on upstream fixes.

lukaszgemborowski:
Ok, I'll create a separate PR for the lambda captures. I'll also try to upstream the doctest change to their repository and then we can come back with a back-merge to cereal. The only thing left will be dependency to boost::format in the performance.cpp. I'm not feeling comfortable with changing anything in boost. :-)

lukaszgemborowski:
I've opened the issue against doctest project: https://github.com/onqtam/doctest/issues/357 let's see what happens. I'll close this PR for now and update the #638 just to track things.

PR #640, opened 2020-04-22
[cpp20] explicitly capture 'this' as copy

lukaszgemborowski:
separate PR for lambda captures created as requested in #639


AzothAmmo:
Will accept after tests pass

Issue #582, opened 2019-09-26
Type registration would not compile with clang on Windows


boguscoder:
Minimal repro:
```
#include <cereal/types/polymorphic.hpp>

struct foo{
    virtual ~foo() = default;
};

CEREAL_REGISTER_TYPE_WITH_NAME(foo, "foo");

int main() {
    return 0;
}
```
Gets

> cereal/details/static_object.hpp:67:29: error:
>       'cereal::detail::StaticObject<cereal::detail::bind_to_archives<foo, cereal::detail::(anonymous
>       namespace)::polymorphic_binding_tag> >' must have external linkage when declared 'dllexport'
>     class CEREAL_DLL_EXPORT StaticObject
>                             ^
> foo.cpp:7:1: note: in instantiation of template class
>       'cereal::detail::StaticObject<cereal::detail::bind_to_archives<foo, cereal::detail::(anonymous
>       namespace)::polymorphic_binding_tag> >' requested here
> CEREAL_REGISTER_TYPE_WITH_NAME(foo, "foo");
> ^

This is caused by default polymorphic tag being in anonymous namespace, making it non-anon would fix the issue

---

EDIT: this is likely that https://bugs.llvm.org/show_bug.cgi?id=43471

InBetweenNames:
Related: #595

AzothAmmo:
reportedly fixed with #643

PR #643, opened 2020-05-27
Fixes to prevent clang-diagnostic errors

johngladp:
When building projects with Visual Studio and cmake and in addition using clang-tidy we see clang-diagnostic errors in the cereal codebase during the clang-tidy execution.  The binaries are built using the msvc compiler (visual studio 16.6.0, Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28805 for x64), successfully.  clang-tidy uses the llvm tool chain and thus the clang compiler (version 9.0.0).

This scenario occurs during the clang-tidy run.  clang-tidy is executed using the build commands generated by cmake, as such _MSC_VER is defined even though the clang compiler is being utilised.

The following diagnostics are generated during the clang-tidy run:

C:\Src\****************\out\build\Windows-x64-Debug\_deps\cereal-src\package\include\cereal\details\static_object.hpp(67,29): error G70D0A859: 'cereal::detail::StaticObject<cereal::detail::bind_to_archives<datapath::udp_stream::send_session_configuration, cereal::detail::(anonymous namespace)::polymorphic_binding_tag> >' must have external linkage when declared 'dllexport' [clang-diagnostic-error]
      class CEREAL_DLL_EXPORT StaticObject
                              ^

And...

C:\Src\******************\out\build\Windows-x64-Debug\_deps\cereal-src\package\include\cereal\details\polymorphic_impl.hpp(658,91): error GD8FB81D5: unexpected type name 'type': expected expression [clang-diagnostic-error]
              savePolymorphicSharedPtr( ar, ptr, ::cereal::traits::has_shared_from_this<T>::type() ); // MSVC doesn't like typename here
......

A guard is required to prevent the un-versioned _MSC_VER test from being taken when compiling with clang under these circumstances.  The clang compiler identifies itself using the __clang__ pre-defined macro which can be used to provide a guard.

### DIFF
--- a/include/cereal/details/polymorphic_impl.hpp
+++ b/include/cereal/details/polymorphic_impl.hpp
@@ -654,7 +654,7 @@ namespace cereal
 
             auto ptr = PolymorphicCasters::template downcast<T>( dptr, baseInfo );
 
-            #ifdef _MSC_VER
+            #if defined(_MSC_VER) & !defined(__clang__)
             savePolymorphicSharedPtr( ar, ptr, ::cereal::traits::has_shared_from_this<T>::type() ); // MSVC doesn't like typename here
             #else // not _MSC_VER
             savePolymorphicSharedPtr( ar, ptr, typename ::cereal::traits::has_shared_from_this<T>::type() );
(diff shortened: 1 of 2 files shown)


johnkearney:
This would also fix:
 - https://github.com/USCiLab/cereal/issues/652
 - https://github.com/USCiLab/cereal/issues/582

PR #646, opened 2020-05-31
Add Bazel build files

tomtseng:
This PR adds a Bazel build file for cereal headers.

For example, if we put the code sample from the cereal README in a file called `my_program.cc` and add the Bazel rule 
```
cc_binary(
    name = "my_program",
    srcs = ["my_program.cc"],
    deps = ["//:cereal"],
)
```
then the code sample compiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@
 *.idb
 *\build_*
 
+# Bazel build output
+/bazel-*
+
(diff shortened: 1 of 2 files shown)


tomtseng:
Wrong fork. I meant to open this PR against `bazelregistry/cereal`.

PR #657, opened 2020-10-20
C++17: use inline globals for StaticObjects

### LINKED ISSUE #582: Type registration would not compile with clang on Windows

### LINKED ISSUE #652: cereal compilation failure VS2019 with clang
Compiles fine using msvc, doesn't with clang. Any ideas?

```
Severity	Code	Description	Project	File	Line	Suppression State
Error		'cereal::detail::StaticObject<cereal::detail::bind_to_archives<TestEntity, cereal::detail::(anonymous namespace)::polymorphic_binding_tag> >' must have external linkage when declared 'dllexport'	C:\Ben\Blah\out\build\x64-Clang-Debug\Floe	C:\Ben\Blah\lib\cereal\include\cereal\details\static_object.hpp	67	
```

```
#include "cereal/types/polymorphic.hpp"
class TestEntity : public Entity
{
...
	template<class Archive>
	void save( Archive& archive ) const
	{
...
	}
	template<class Archive>
	void load( Archive& archive )
	{
...
	}
};

CEREAL_REGISTER_TYPE( TestEntity )
CEREAL_REGISTER_POLYMORPHIC_RELATION( Entity, TestEntity )
```

InBetweenNames:
This prevents multiple definition errors in Clang,
and also stops dllexporting functions with internal
linkage.  Degrades gracefully when C++17 is not
present.  Tested on MSVC 2017 and LLVM on Windows.
Also fix a few `_MSC_VER` tests to only apply for non-recent
versions of MSVC 2017 and below.

Fix #595
Fix #652
Fix #582
Fix #643

### DIFF
--- a/include/cereal/archives/json.hpp
+++ b/include/cereal/archives/json.hpp
@@ -286,7 +286,7 @@ namespace cereal
       void saveLong(T lu){ saveValue( static_cast<std::uint64_t>( lu ) ); }
 
     public:
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && _MSC_VER < 1916
       //! MSVC only long overload to current node
       void saveValue( unsigned long lu ){ saveLong( lu ); };
 #else // _MSC_VER
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -258,6 +258,26 @@ namespace cereal
       Interfaces for other forms of serialization functions is similar.  This
       macro should be placed at global scope.
       @ingroup Utility */
+
+  //! On C++17, define the StaticObject as inline to merge the definitions across TUs
+  //! This prevents multiple definition errors when this macro appears in a header file
+  //! included in multiple TUs.
+  #ifdef CEREAL_HAS_CPP17
+  #define CEREAL_CLASS_VERSION(TYPE, VERSION_NUMBER)                             \
+  namespace cereal { namespace detail {                                          \
+    template <> struct Version<TYPE>                                             \
+    {                                                                            \
+      static std::uint32_t registerVersion()                                     \
+      {                                                                          \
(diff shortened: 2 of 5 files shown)


InBetweenNames:
Made a small mistake with `#if`, now correctly checks for `defined(_MSC_VER)` instead of blindly comparing the version number.

InBetweenNames:
Interesting.  It seems like Clang 5 has a bug with inline variables that was fixed in later versions.  It seems that there are two ways forward: we could remove the test or we could adjust the `CEREAL_HAS_CPP17` macro to specifically exclude Clang 5.  What would be the best way to go forwards?

InBetweenNames:
Fix applied to exclude Clang 5.x from C++17 features

breznak:
Very nice fix, thank you! Any chance to get this merged soon?

Issue #459, opened 2017-12-20
optional fields when serialize and deserialize [json]


xoac:
I have tested develop branch that supports `std::optional` This works like this:  
C++:      
`std::optional<bool> test = std::make_optional(true);`
json:  
```
"test": {
        "nullopt": false,
        "data": true
    }
```

what is more this optional require from me to provide this that if fields not exist. (when reading json)

Is some recommend way to do below in cereal? 

1. If `std::optional` doesn't contain elements serializer doesn't generate anything in json.
2. if `std::optional` contains data this generate filled with that data. Taking above example it would be:
```
"test": true,
```
3. Loading optional data, no exception when trying load key doesn't exist and std::optional stay empty or unchanged. 

I have found some information there was planning for function like `cereal::make_optional` etc. to service not required fields, but nothing like this in library at the moment.

AzothAmmo:
You can specialize any of the serialization functions cereal ships with for any specific archive or type. In addition, you can write custom prologue/epilogue functions (these get called right before and right after serialization) in a way that has a custom interaction with an archive.

The discussion in #30 will be most relevant to actually trying to implement this. Currently the json archive will throw an exception if an expected NVP does not exist - you can work around this by catching this exception and treating it as `nullopt` in your specialized serialization code.

xoac:
There are proposed solutions that both are incorrect.  
* First https://github.com/Enhex/Cereal-Optional-NVP required order.
* Second (Exception) https://github.com/LowCostCustoms/cereal-optional has a big flue, that if the optional filed exist but for example has improper data (expect bool but contains int) wouldn't be load. 

The closed in my opinion solution is https://github.com/USCiLab/cereal/issues/30#issuecomment-223825543 but there in no access to `archiver.itsIteratorStack.back()` I would like to increment it if `getNodeName()` wouldn't contain a expected value.

xoac:
**So, for now it's impossible to check `filed` exist or not?**

xoac:
Closing because of inactivity. Feel free to reopen

PR #666, opened 2020-12-19
Eliminate JSON entry for empty std::optional

lowsfer:
This change remove the entry in JSON if what is being serialized/deserialized is of type std::optional<T>, and the value is std::nullopt. When it's not std::nullopt, only the contained value (T) is saved, instead of a flag plus the value.

This is helpful when using JSON as web protocol. People would want to eliminate optional entries when they are empty. This improved human-readability of the JSON protocol and make it easier to interact with other languages.


lowsfer:
Seems this is exactly what is requested here:
https://github.com/USCiLab/cereal/issues/459

WSoptics:
It might be a good idea to make this an option of `JSONArchive` instead of altering its default behavior?

lowsfer:
@WSoptics Changed it to an option and defaults to false. Also added unit tests.

fiesh:
I like the change but can't merge :-)

One thing though however, it seems the option is only available for C++17, which seems not really necessary?

You should probably ping @AzothAmmo to potentially have it merged.

lowsfer:
@fiesh std::optional is only available since C++17. You mean keep the option but leave it with no effect?

@AzothAmmo Can we merge this feature?

fiesh:
> @fiesh std::optional is only available since C++17. You mean keep the option but leave it with no effect?

Sorry, I somehow thought `std::optional` was C++11 -- you're right of course!

winding-lines:
Would also love to have this behavior in cereal, @AzothAmmo :)

lowsfer:
It has been ~1.75 years.
@AzothAmmo Do you have any concerns on this PR?

akalsi87:
This seems like a good change, and one that makes the output more intuitive and interoperable.  Any reason for the delay in merging this in?

Yu-zqi:
@lowsfer it seems this change can't support out-of-order json fileds.

Yu-zqi:
I made a few changes to support out-of-order JSON fields. It works fine in my case, but I don't know if there are any side effects.

here are my changes:
```c++
template <class T> inline  std::enable_if_t<IsOptional<T>::value>
CEREAL_LOAD_FUNCTION_NAME( JSONInputArchive & ar, NameValuePair<T> & t )
{
  if (ar.skippedNullopt())
  {
    // search node name
    if (ar.searchNodeName(t.name))
    {
      ar.setNextName( t.name );
      std::decay_t<decltype(*t.value)> val;
      ar( val );
      t.value = std::move(val);
    }
    else
    {
      t.value = std::nullopt;
    }
  }
  else
  {
    loadImpl(ar, t);
  }
}
``` 

```c++
const char* searchNodeName(const char* name) const
{
  auto currentIterator = itsIteratorStack.back();
  const auto len = std::strlen( name );
  for (const auto& it : currentIterator)
  {
    const auto currentName = it.name.GetString();
    if( ( std::strncmp( name, currentName, len ) == 0 ) &&
        ( std::strlen( currentName ) == len ) )
    {
    return currentName;
    }
  }
  return nullptr;
}
```

```c++
// Iterator support range for syntax
auto begin()
{
  return itsMemberItBegin;
}

auto end()
{
   return itsMemberItEnd;
}
```

lowsfer:
@yuziqii Would you like to create a commit including your changes? You can create an PR into my forked branch lowsfer:eliminate-empty-optional, then I merge it and your commit will be included in this PR.

yuziqii:
@lowsfer Sure, I'd be happy to submit a PR , and I've already done that.  Let me know if there's any problems. here is the [PR](https://github.com/lowsfer/cereal/pull/1).

PR #667, opened 2020-12-19
CVE-2020-11105: Store a copy of each serialized shared_ptr within the archive to prevent the shared_ptr to be freed to early.

### LINKED ISSUE #636: std::shared_ptr serialization asymmetry (depends on memory layout)
Cereal employs caching of ```std::shared_ptr``` values, using the raw pointer as a unique identifier. This becomes problematic if an ```std::shared_ptr``` variable goes out of scope and is freed, and a new ```std::shared_ptr``` is allocated at the same address. Serialization fidelity thereby becomes dependent upon memory layout.

```cpp
#include <cereal/archives/binary.hpp>
#include <cereal/types/memory.hpp>

int main(void)
{
    std::stringstream ss;

    {
        cereal::BinaryOutputArchive archiveOut(ss);
        {
            std::shared_ptr<bool> v = std::make_shared<bool>(true);
            archiveOut(v);
            printf("v is %p\n", v.get());
            printf("serialized: %d\n", *v);
        }
        {
            std::shared_ptr<bool> v = std::make_shared<bool>(false);
            archiveOut(v);
            printf("v is %p\n", v.get());
            printf("serialized: %d\n", *v);
        }
    }

    {
        cereal::BinaryInputArchive archiveIn(ss);
        std::shared_ptr<bool> v1, v2;
        archiveIn(v1);
        printf("deserialized: %d\n", *v1);
        archiveIn(v2);
        printf("deserialized: %d\n", *v2);
    }
    return 0;
}
```

Output is:

```
v is 0x5578c0144ec0
serialized: 1
v is 0x5578c0144ec0
serialized: 0
deserialized: 1
deserialized: 1
```

The input is (true, false) but the output is (true, true).

serpedon:
The archives use the memory address pointed by the shared_ptr as a
unique id which must not be reused during lifetime of the archive.
Therefore, the archives stores a copy of it.
This problem was also reported as CVE-2020-11105 and #636. This should fix #636.

My line of though was the following:
Correct usage of std::shared_ptr in Cereal requires that the shared_ptr is still valid at the point when all serialization occurs, usually at the end of the lifetime of the archive. It was suggested to document this constraint to the user, but since we are already dealing with smart pointers, I though, hey, let's implement this constraint by storing our own copy of the std::shared_ptr.

### DIFF
--- a/include/cereal/cereal.hpp
+++ b/include/cereal/cereal.hpp
@@ -369,12 +369,17 @@ namespace cereal
           point to the same data.
 
           @internal
-          @param addr The address (see shared_ptr get()) pointed to by the shared pointer
+          @param sharedPointer The shared pointer itself (the adress is taked via get()).
+                               The archive takes a copy to prevent the memory location to be freed
+                               as long as the address is used as id. This is needed to prevent CVE-2020-11105.
           @return A key that uniquely identifies the pointer */
-      inline std::uint32_t registerSharedPointer( void const * addr )
+      inline std::uint32_t registerSharedPointer(const std::shared_ptr<const void>& sharedPointer)
       {
+        void const * addr = sharedPointer.get();
+
         // Handle null pointers by just returning 0
         if(addr == 0) return 0;
+        itsSharedPointerStorage.push_back(sharedPointer);
 
         auto id = itsSharedPointerMap.find( addr );
         if( id == itsSharedPointerMap.end() )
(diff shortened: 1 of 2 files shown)


AzothAmmo:
Looks good, I'm trying to figure out why the Travis CI did not run - will merge after that is sorted.

InBetweenNames:
Looks like all checks have passed -- can we get this merged?

PR #674, opened 2021-01-31
feature: named variants in json serialization

miartad:
Hello, 

I made some changes to how variants are represented in json for my use case and putting it here in case you would also find it useful.

example output:

1) g++ --std=c++17 m.cpp
```
{
    "value0": {
        "variant": {
            "index": 0,
            "data": {
                "id": 5
            }
        }
    }
}
```

2) g++ -DCEREAL_NAMED_VARIANT --std=c++17 m.cpp
```
{
    "value0": {
        "variant": {
            "type": "LoadInstrument",
            "LoadInstrument": {
                "id": 5
            }
        }
    }
}
```
code example:
https://gist.github.com/miartad/3b290142284a4bed0a8c1736d9cc62ed

### DIFF
--- a/include/cereal/types/variant.hpp
+++ b/include/cereal/types/variant.hpp
@@ -33,6 +33,9 @@
 #include "cereal/cereal.hpp"
 #include <variant>
 #include <cstdint>
+#ifdef CEREAL_NAMED_VARIANT
+#include <map>
+#endif
 
 namespace cereal
 {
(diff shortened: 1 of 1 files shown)


WSoptics:
As discussed in another pull request, this is problematic since `std::variant` allows for the same type to exist multiple times.  So this makes no sense in general.

miartad:
are variants with multiples of same types supported by cereal? in my test these types of variants do not compile with cereal.

PR #675, opened 2021-01-31
feature: represent char arrays as strings in json

miartad:
Hello, 

This feature allows one dimensional array to be represented as string in json.

example output:

1) g++ --std=c++17 m.cpp
```
{
    "value0": {
        "name": {
            "value0": 97,
            "value1": 100,
            "value2": 97,
            "value3": 109,
            "value4": 0,
            "value5": 0,
            "value6": 0,
            "value7": 0
        }
    }
}
```

2) g++ -DCEREAL_NAMED_VARIANT --std=c++17 m.cpp
```
{
    "value0": {
        "name": "adam"
    }
}
```

multidimensional array would look like:
```
{
    "value0": {
        "name": {
            "value0": "adam",
            "value1": "mada"
        }
    }
}
```
code example:
https://gist.github.com/miartad/83ff9033da1080e55103e12e7763c772

### DIFF
--- a/include/cereal/types/common.hpp
+++ b/include/cereal/types/common.hpp
@@ -32,6 +32,8 @@
 
 #include "cereal/cereal.hpp"
 
+#include <cstring>
+
 namespace cereal
 {
   namespace common_detail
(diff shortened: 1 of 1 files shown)


WSoptics:
What's the point of not using a `std::string` to begin with?

I think this pull request should be rejected since:
* It changes global behavior based on a preprocessor directive.
* Its benefit is dubious.

miartad:
In my use case I'm using fixed size char arrays to represent strings because std::string allocates on heap. I guess it is a niche use case and will be deprecated when c++ std has some kind of support for stack allocated strings, which I dont think it currently has.

WSoptics:
> In my use case I'm using fixed size char arrays to represent strings because std::string allocates heap. I guess it is a niche use case and will be deprecated when c++ std has some kind of support for stack allocated strings, which I dont think it currently has.

C++ does support this already through polymorphic memory resources, see `std::pmr::string`.  So I guess the correct approach would to support those natively in cereal.

StefanBruens:
std::string is stack-only for short strings (SSO).
If used in a loop, put the string outside the loop and clear it once per iteration, this removes most heap allocations from the loop body.

std::string_view can be put on arbitrary binary storage.

PR #705, opened 2021-07-20
add vector child name

Benjamin-Tan:


Benjamin-Tan:
#400 rename from withInnerName to withVectorChildName

PR #716, opened 2021-09-21
Update doctest to 2.4.6 + local fixes slated for upstream

AzothAmmo:


AzothAmmo:
Still trying to figure out a few issues with this upstream, follow along here: https://github.com/AzothAmmo/doctest/tree/noreturn_macro

PR #717, opened 2021-09-26
Fixed compile error related to std::string in JSONOutputArchive

Bouloulou1138:
JSONOutputArchive expected an std::string param to be passed saveValue().
This caused the compilation to fail when the more generic std::basic_string was passed and not an std::string.
This fix changes the signature of saveValue to expect an std::basic_string


Bouloulou1138:
Hello @AzothAmmo, could this simple fix be added to the v1.3.1 release?

Boulougou:
I am also affected by this compilation issue, would this fix be included in the next release?

PR #725, opened 2021-11-27
Support josn long long linux (ubuntu/android) gcc/clang

crazyhappygame:
Problem:
json serialization fails with 
`'terminating with uncaught exception of type cereal::RapidJSONException: rapidjson internal assertion failure: IsString()'`


crazyhappygame:
Closing PR for now.
1. Ci does not work
2. there are problems with other types

PR #739, opened 2022-01-17
build just the tests, as an option for Debian autopkgtests later

mr-c:
With assistance from @kdmurray91; thanks!


mr-c:
We've been carrying this patch in Debian since 2016, so I thought I'd submit it :-)

https://salsa.debian.org/med-team/libcereal/-/blob/master/debian/patches/add-only-tests-target

AzothAmmo:
addressing this through #751

mr-c:
Thanks!

PR #758, opened 2022-07-03
Fix a performance bug when deserializing C arrays

zainryan:
Without the proposed change, cereal will deserialize the C-array using the slow-path serializeArray(), i.e., the one that deserializes the array element by element, even though the array is arithmetic.


zainryan:
Can anyone have a quick review on this? It makes a big performance difference in my use case.

AzothAmmo:
I think the proper fix for this would for us to define separate load and save functions checking either `is_input_serializable` or `is_output_serializable`. For your use case, it is on that this is not symmetric and makes me think there is a deeper issue either in your code or something wrong in cereal.

PR #761, opened 2022-08-10
fix defer variable corruption

rajashekar007:
As deferements are iterated using range based loop, when there is a recursive defer, it's corrupting the variable. The loop iteration has been changed to use index based to fix the same.

Kindly review it.


AzothAmmo:
This strikes me as likely an incorrect usage of `cereal::defer` - it shouldn't be used in situations where there is any recursive structure to serialize as this defeats the purpose of the deferment in the first place.

PR #775, opened 2023-01-02
Disable unnecessary stuff if not main project

LucaCiucci:
I had some problems while including cereal in my project using `FetchContent`, so I made some small modifications to the `CMakeLists.txt`:
 - renamed some options using the `CEREAL_` prefix to make clear to which project they belong to and avoid conflicts
 - changed the default value of some options so that, when not the main project, doc, sandbox and performance comparisons are disabled by default
 - added a more complete check for main project

I think this edits would make it easier to depend on cereal using `FetchContent`, including it as submodule or similar techniques. See [this](https://github.com/SpartanJ/efsw/pull/149) other pull request for more info.


LucaCiucci:
I see that some CI checks are not successful, but the errors are unrelated to this particular PR, am I right?

AzothAmmo:
Yes CI seems to be borked - was hoping to go through and slam through a bunch of these PRs but will have to fix that before I do so.

PR #792, opened 2023-06-29
Use CTTI instead of RTTI where possible

sharkov63:
In some projects we are required to use `-fno-rtti` option, which disables RTTI (run-time type information). `cereal` uses [typeid](https://en.cppreference.com/w/cpp/language/typeid) operator, which requires RTTI, in the following two scenarios:
1. using `typeid` on a type which is known at compile-time (usually taking `hash_code()` of the returned `std::type_info`);
2. using `typeid` in order to support serialization of polymorphic types (i.e. here types are not known at compile time), see [polymorphic.hpp](https://github.com/USCiLab/cereal/blob/master/include/cereal/types/polymorphic.hpp).

While usage in the scenario 2 is required to serialize polymorphic types, we don't need RTTI at all in scenario 1.

This PR replaced all usages of `typeid` having a compile-time known type with methods from an external library [ctti](https://github.com/Manu343726/ctti) (open source, MIT license) - this library provides `ctti::type_id<T>()` functions for compile-time type id, and also `ctti::type_id_t` type, similar to `std::type_info`.

With this change one can use cereal with `-fno-rtti` flag, in case when one does not serialize any polymorphic types (i.e. does not include `polymorphic.hpp` or other related headers).


sharkov63:
Converted to draft because there is at least another instance of using `typeid` with compile-time known type: see `util.hpp`

PR #809, opened 2023-12-10
std::filesystem::path support

f-michaut:
Adding `std::filesystem::path` support in cereal.

Relying on the current `std::string` serialization since it is the underling storage for filesystem::path.


AzothAmmo:
This looks pretty good, it needs a unit test like the other types, and that unit test should be properly gated behind C++17 conditionals like similar features.

PR #812, opened 2024-01-11
Remove template parameter `Archive` from `InputBindingsMap` and `OutputBindingsMap`

arekfu:
This PR is a rebase of #521. I am not the original author of the code in this PR (credit: @uentity), but I am willing to push it through integration.


AzothAmmo:
Thank you for the rebase - the only thing holding this up and holding up most PRs is that I haven't gotten around to fixing the various CI pipelines, which inexplicably broke a while back.

PR #816, opened 2024-02-29
More robust __cplusplus check

dimateos:
Similar to the same check done a few lines above for c++17


redchairman:
您的邮件我已收到，谢谢合作！

PR #826, opened 2024-07-21
Reduced string operations when searching for JSON node names

EmilianC:
The first `strlen()` call required that the `searchName` is iterated entirely, even if the following `strncmp()` would be able to early-out when the first characters don't match.

The pre-calculated length is only used to constrain the string comparison, but `strcmp()` will handle the ends of both strings just fine, and early-out as soon as possible.

The second call to `strlen()` was to ensure that `searchName` didn't just match a prefix of a longer `currentName`. `strcmp()` also accounts for this, but will early-out if the string is a single character longer, without having to walk the entire `currentName`.


EmilianC:
Pardon, silly extra `, len` argument snuck in last minute when making the PR :) Fixed in [7d1d4c5](https://github.com/USCiLab/cereal/commit/7d1d4c54049c5363cc6e518cbae02f93fdeb72a0)

PR #835, opened 2024-09-11
Fix instances of -Wmissing-template-arg-list-after-template-kw.

kraj:
Clang has a new warning that requires a template argument list after using the template keyword. Remove uses of the template keyword when we're not specifying types.

See https://github.com/llvm/llvm-project/issues/94194 for the upstream clang changes


christoph-cullmann:
Would be nice to have that to fix the warning.

PR #846, opened 2024-12-09
Fix linux CI

sjrdc:
Use docker-in-docker construction to run old (unsupported) containers for old GCC versions.
MacOS builds are still failing, as are windows builds.


AzothAmmo:
testing now, will rebase and merge if everything looks good

AzothAmmo:
merged

PR #848, opened 2024-12-23
Fix `std::array<T,0>`

PeizeLin:
For `std::array<T,0>`, the `sizeof(array)!=0`.
Then bugs appear at
https://github.com/USCiLab/cereal/blob/master/include/cereal/types/array.hpp#L45
and
https://github.com/USCiLab/cereal/blob/master/include/cereal/types/array.hpp#L55 ,
since `array.data()` is `nullptr` for `std::array<T,0>`.

### DIFF
--- a/include/cereal/types/array.hpp
+++ b/include/cereal/types/array.hpp
@@ -42,13 +42,8 @@ namespace cereal
                           && std::is_arithmetic<T>::value, void>::type
   CEREAL_SAVE_FUNCTION_NAME( Archive & ar, std::array<T, N> const & array )
   {
-    ar( binary_data( array.data(), sizeof(array) ) );
+    ar( binary_data( array.data(), N*sizeof(T) ) );
   }
-
-  template <class Archive, class T> inline
-  typename std::enable_if<traits::is_output_serializable<BinaryData<T>, Archive>::value
-                          && std::is_arithmetic<T>::value, void>::type
-  CEREAL_SAVE_FUNCTION_NAME( Archive & ar, std::array<T, 0> const & array ) {}
 
   //! Loading for std::array primitive types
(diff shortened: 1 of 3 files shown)


AzothAmmo:
Do you mind adding a test case to `https://github.com/USCiLab/cereal/blob/master/unittests/array.hpp`?

PeizeLin:
> Do you mind adding a test case to `https://github.com/USCiLab/cereal/blob/master/unittests/array.hpp`?

OK, I've added the test cases.

sjrdc:
I would expect that the same kind of error occurs for valarray of size 0, vector of size 0, etc. Shouldn't you fix this in the handling of binary_data created of nullptr and size 0 by the archive class, rather than the specific type(s)? That way we have a generic solution for the situation.

PeizeLin:
> I would expect that the same kind of error occurs for valarray of size 0, vector of size 0, etc. Shouldn't you fix this in the handling of binary_data created of nullptr and size 0 by the archive class, rather than the specific type(s)? That way we have a generic solution for the situation.

I've checked that vector and valarray will not occur error.
It can be seen from
https://github.com/USCiLab/cereal/blob/master/include/cereal/types/vector.hpp#L45
https://github.com/USCiLab/cereal/blob/master/include/cereal/types/valarray.hpp#L47
The binary_data size is vector.size(), which is indeed 0.

But, perhaps we can change
`ar( binary_data( array.data(), sizeof(array) ) );`
to
`ar( binary_data( array.data(), N*sizeof(T) ) );`
@AzothAmmo Maybe this is a more generic solution?

sjrdc:
Great suggestion, or use data.size()

PeizeLin:
> Great suggestion, or use array.size()

updated

PR #866, opened 2025-07-04
Update boost path in appveyor CI

stephematician:
Should be a relatively straightforward update to the Boost paths so that appveyor builds succeed.


redchairman:
您的邮件我已收到，谢谢合作！

AzothAmmo:
re-running CI but looks good

PR #872, opened 2025-09-14
Fix llvm20 compile issue in tuple.hpp

nindanaoto:
I encountered a compilation issue with my library using cereal when I tried to compile it with clang++-20.  GCC is still working fine, but it is not accepted by clang++-20. This PR introduces a minor change that fixes the issue. AFAIK, GCC is accepting this because of the bug. 
https://github.com/llvm/llvm-project/issues/94194


jarzec (approved):
LGTM
Explicit template specialization seems to be slightly safer and expresses intent better.

Issue #865, opened 2025-05-12
Original order of keys lost after deserialization for std::multimap and std::multiset


lordofdestiny:
## Description
As defined per the C++11 standard, as referenced her for [std::multimap](https://en.cppreference.com/w/cpp/container/multimap#:~:text=The%20order%20of%20the%20key%2Dvalue%20pairs%20whose%20keys%20compare%20equivalent%20is%20the%20order%20of%20insertion%20and%20does%20not%20change.) and [std::multiset](https://en.cppreference.com/w/cpp/container/multiset#:~:text=The%20order%20of%20the%20elements%20that%20compare%20equivalent%20is%20the%20order%20of%20insertion%20and%20does%20not%20change.), 
The order of the elements/keys-value pairs that compare equivalent is the order of insertion and does not change. There is a bug in how the deserialization of these standard containers is implemented, which results in order of equivalent keys being flipped after each subsequent deserialization. 

## How to reproduce
Here is a minimal code to reproduce the bug:
```c++
#include <iostream>
#include <string>
#include <iterator>
#include <sstream>

#include "cereal/types/map.hpp"
#include "cereal/types/set.hpp"
#include "cereal/archives/binary.hpp"

struct MockValue
{
    char key, value;

    MockValue() = default;

    explicit MockValue(char key, char value)
        : key(key), value(value) {}

    bool operator<(const MockValue &other) const
    {
        return this->key < other.key;
    }
    bool operator==(const MockValue &other) const
    {
        return this->key == other.key && this->value == other.value;
    }

    friend std::ostream &operator<<(std::ostream &os, const MockValue &val)
    {
        return os << "[Key=" << val.key << ", Value=" << val.value << "]";
    }

    template <class Archive>
    void serialize(Archive &archive)
    {
        archive(key, value);
    }
};

// Print generic std::pair
template <template <typename...> class Pair, typename... Args,
          typename = std::pair<typename Pair<Args...>::first_type,
                               typename Pair<Args...>::second_type>>
std::ostream &operator<<(std::ostream &os, const Pair<Args...> &pair)
{
    return os << "Pair{" << pair.first << ", " << pair.second << "}";
}

// Print standard containers
template <
    template <typename...> class Container,
    typename... Args>
void printContainer(Container<Args...> &container)
{
    using VType = typename Container<Args...>::value_type;
    auto begin = container.begin();
    auto end = container.end();
    std::copy(begin, end, std::ostream_iterator<VType>(std::cout, "\n"));
[...]

PR #874, opened 2026-01-22
Fix multicontainer deserialization

lordofdestiny:
# Subject

As reported in this issue ( https://github.com/USCiLab/cereal/issues/865 ), deserialization of the std::multimap and std::multiset breaks the container invariants. These containers are expected ( by the standard ), to maintain the relative (insertion) order of the values who's keys evaluate as equivalent. 

The `load` method for these containers uses `insert_hint(hint, value)` ( or `emplace_hint` ) methods to add values to the container during load. This results in reversing the order of the elements with equivalent keys, because the method places the new element before the `hint` instead of after. 

# Solution

Implement handlers for these method separately instead of generically. Regular emplace is needed in order to maintain the insertion order of equivalent keys. For keys that are not equivalent there is no expectation of order so they can be placed in any order. In order to still maintain the benefits of the faster `emplace_hint`, it is still used when the current key is not equivalent to the previous. If the current key is equivalent to the previous key, `emplace` needs to be used.

# Testing

By running the following cmake commands old I asserted that the current behaviour was not broken:
```bash
> cmake -DCMAKE_BUILD_TYPE=Debug -DSKIP_PERFORMANCE_COMPARISON=YES -B build -DSKIP_PORTABILITY_TEST=TRUE -DCMAKE_CXX_STANDARD=17
> cmake --build build
> cmake --build build --target test
```

The new behaviour was tested using the script given in the original issue.

# Compilation issues
The cmake build seems to be broken in some files. In order to build the test at all the following diffs were applied:

```diff
diff --git a/include/cereal/details/helpers.hpp b/include/cereal/details/helpers.hpp
index 8dafb8e..0f058ac 100644
--- a/include/cereal/details/helpers.hpp
+++ b/include/cereal/details/helpers.hpp
@@ -244,8 +244,6 @@ namespace cereal
       static_assert( !std::is_base_of<detail::DeferredDataCore, T>::value,
                      "Cannot defer DeferredData" );
 
-      DeferredData & operator=( DeferredData const & ) = delete;
-
     public:
       //! Constructs a new NameValuePair
       /*! @param v The value to defer.  Ideally this should be an l-value reference so that
```

```diff
diff --git a/include/cereal/types/tuple.hpp b/include/cereal/types/tuple.hpp
index 80c6807..d690199 100644
--- a/include/cereal/types/tuple.hpp
+++ b/include/cereal/types/tuple.hpp
@@ -95,7 +95,7 @@ namespace cereal
       template <class Archive, class ... Types> inline
       static void apply( Archive & ar, std::tuple<Types...> & tuple )
       {
-        serialize<Height - 1>::template apply( ar, tuple );
+        serialize<Height - 1>::apply( ar, tuple );
         ar( CEREAL_NVP_(tuple_element_name<Height - 1>::c_str(),
             std::get<Height - 1>( tuple )) );
       }
@@ -116,7 +116,7 @@ namespace cereal
   template <class Archive, class ... Types> inline
   void CEREAL_SERIALIZE_FUNCTION_NAME( Archive & ar, std::tuple<Types...> & tuple )
   {
-    tuple_detail::serialize<std::tuple_size<std::tuple<Types...>>::value>::template apply( ar, tuple );
+    tuple_detail::serialize<std::tuple_size<std::tuple<Types...>>::value>:: apply( ar, tuple );
   }
 } // namespace cereal
 ```
 
 ```diff
 diff --git a/unittests/doctest.h b/unittests/doctest.h
index cd5b44d..307315f 100644
--- a/unittests/doctest.h
+++ b/unittests/doctest.h
@@ -3631,7 +3631,7 @@ String toString(double long in) { return fpToString(in, 15); }
 #define DOCTEST_TO_STRING_OVERLOAD(type, fmt)                                                      \
     String toString(type in) {                                                                     \
         char buf[64];                                                                              \
-        std::sprintf(buf, fmt, in);                                                                \
+        std::snprintf(buf, 64, fmt, in);                                                           \
         return buf;                                                                                \
     }
 ```

### DIFF
--- a/include/cereal/types/map.hpp
+++ b/include/cereal/types/map.hpp
@@ -33,4 +33,55 @@
 #include "cereal/types/concepts/pair_associative_container.hpp"
 #include <map>
 
+namespace cereal
+{
+  /**
+    * std::multimap must not default to the implementation in
+    * pair_associative_container.hpp, because it incorrectly
+    * handles equivalent keys when loading the map.
+    *
+    * The cause is the use of emplace_hint.
+    * Since it iserts the value before the hint, key order will be flipped
+    *
+    * Here we are using emplace (if key is equivalent to the previous one),
+    * or emplace_hint if keys are different, because this is a faster path
+    *
+    * Emplace will preserve order of equivalent keys, while where
+    * non-equivalent ones are placed doesn't matter.
+    **/
+
+  template <class Archive, typename... Args>
+  inline void load(Archive &ar, std::multimap<Args...> &map)
+  {
+    size_type size;
+    ar(make_size_tag(size));
+
+    map.clear();
+
+    typename std::multimap<Args...>::key_compare cmp;
+    auto hint = map.begin();
+    for (size_t i = 0; i < size; ++i)
+    {
+      typename std::multimap<Args...>::key_type key;
+      typename std::multimap<Args...>::mapped_type value;
+
+      ar(make_map_item(key, value));
+      if(!cmp(hint->first, key) && !cmp(key, hint->first)) {
+        #ifdef CEREAL_OLDER_GCC
+        hint = map.insert(std::move(key), std::move(value));
+        #else // NOT CEREAL_OLDER_GCC
+        hint = map.emplace(std::move(key), std::move(value));
+        #endif // NOT CEREAL_OLDER_GCC
+      } else {
+        #ifdef CEREAL_OLDER_GCC
+        hint = map.insert_hint(hint, std::move(key), std::move(value));
+        #else // NOT CEREAL_OLDER_GCC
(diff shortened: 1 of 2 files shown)


lordofdestiny:
> FWIW, I think all that needs doing is to increment the iterator for the next hint: https://github.com/lukester1975/cereal/commit/116e63d48aca20051b03942c18c18a558ca02596
> 
> Cheers

I say if it works then fine. Didn't really cross my mind at all. As long as I pointed out the bug, I'm happy!

lordofdestiny:
@mclayto That is probably the simplest solution. Honestly that didn't cross my mind either. I would be happy to implement it the bug was not solved already

PR #884, opened 2026-07-06
Update CI to use GCC 13 and Ubuntu 24.04

mosfet80:
Updated CI configuration to use newer compiler versions and Ubuntu containers. Removed deprecated compiler configurations.


redchairman:
您的邮件我已收到，谢谢合作！

PR #887, opened 2026-08-07
Allow changing only JSON output whitespace options

mwoehlke-kitware:
Add a secondary constructor to `JSONOutputArchive::Options` that allows users to change the whitespace options while retaining the default floating point precision.


redchairman:
您的邮件我已收到，谢谢合作！

PR #889, opened 2026-09-13
Fix for C++ 20 error

TrevorCash:
Contains fix for C++ 20 Error:

error: C++20 says that these are ambiguous, even though the second is reversed: [-Werror]
  628 |     assert( e == e_out );


redchairman:
您的邮件我已收到，谢谢合作！

